PR #17, opened 2020-06-19
Regen deps

Cynocracy:
./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh

(because some genreconciler change broke this)


Cynocracy:
wut. Logs look fine

/retest

Cynocracy:
Probably fallout from
https://github.com/knative/pkg/pull/1403

Hopefully serving has an example to crib from :)

Cynocracy:
I'm a bit at a loss. Shouldn't geninjection be regenerating the problematic call?

Cynocracy:
Ha, didn't notice it had been merged.

I might have an actual fix now? TBD

PR #22, opened 2020-06-24
Removes unwanted timeout setting

richterdavid:
Remove setting KIngress timeout per comments from @vagababov and @tcnghia in slack: 
https://knative.slack.com/archives/CA9RHBGJX/p1592591518025100

PR #39, opened 2020-07-21
Revert "Removes unwanted timeout setting"

mattmoor:
Reverts knative/networking#22

This is a breaking API change that needs to be tracked as such with the appropriate stakeholders looped in.

/assign @tcnghia @richterdavid

### DIFF
--- a/pkg/apis/networking/v1alpha1/ingress_defaults.go
+++ b/pkg/apis/networking/v1alpha1/ingress_defaults.go
@@ -18,8 +18,12 @@ package v1alpha1
 
 import (
 	"context"
+	"time"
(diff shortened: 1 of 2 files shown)


mattmoor:
At a minimum to roll this change forward, we need a kingress conformance test for the expected behavior of an unspecified timeout, but I think it should get an "Ack" from all of the net-* stakeholders (Gloo, Kong, Contour, Kourier, Ambassador).

yanweiguo:
cc @ZhiminXiang

richterdavid:
FYI tcnghia is OOO today.

PR #57, opened 2020-08-01
Move config-network.yaml from knative.dev/serving to knative.dev/networking

tcnghia:
`config-network.yaml` mostly contains feature flags for networking features that are consumed by the network plugins, it makes sense to be in `knative.dev/networking` rather than `knative.dev/serving`. This move also makes it easier for the networking plugin to avoid depending on knative/serving while still can use an upstream version of `config-network.yaml`.

### DIFF
--- a/config/config-network.yaml
+++ b/config/config-network.yaml
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: "b22469ec"
+    knative.dev/example-checksum: "3317ccca"
 data:
   _example: |
     ################################
@@ -37,13 +37,6 @@ data:
     # These sample configuration options may be copied out of
     # this example block and unindented to be in the data block
     # to actually change the configuration.
-
-    # DEPRECATED:
-    # istio.sidecar.includeOutboundIPRanges is obsolete.
-    # The current versions have outbound network access enabled by default.
-    # If you need this option for some reason, please use global.proxy.includeIPRanges in Istio.
-    #
-    # istio.sidecar.includeOutboundIPRanges: "*"
 
     # ingress.class specifies the default ingress class
     # to use when not dictated by Route annotation.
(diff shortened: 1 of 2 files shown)


tcnghia:
```
diff: /home/prow/go/src/knative.dev/networking/tmp.UcyTTvgEJn/pkg/testdata/config-network.yaml: No such file or directory
ERROR: /home/prow/go/src/knative.dev/networking is out of date. Please run ./hack/update-codegen.sh
2020/08/01 16:28:20 Getting artifacts dir from prow
```

Seems changing file into symlink doesn't work with `verify-codegen.sh`

tcnghia:
Fixed verify-codegen.sh to also copy the `config/` dir.

mattmoor:
/hold

What is our story for consuming this downstream?

mattmoor:
Have we moved other things like the CRD definitions for kingress, kcert?

tcnghia:
this is now too close to release, so we will move yamls in the early weeks of next release.

config/config-network.yaml
```diff
@@ -0,0 +1,118 @@
...
+    # This block is not actually functional configuration,
+    # but serves to illustrate the available configuration
+    # options and document them in a way that is accessible
+    # to users that `kubectl edit` this config map.
+    #
+    # These sample configuration options may be copied out of
+    # this example block and unindented to be in the data block
+    # to actually change the configuration.
+
+    # DEPRECATED:
```
nak3, line 41: I removed this field by https://github.com/knative/networking/pull/108/files#diff-da212c4f484ff918ebc11796bb32a37eL41 so would be great if you could remove it.

nak3:
@tcnghia will you include CRD for kingress and cert? We noticed an issue https://knative.slack.com/archives/C012C0VQJAW/p1599577449008200

tcnghia:
@nak3 let me rebase and check this in first. we can follow up with more from config/ later.

tcnghia:
@googlebot I signed it!

tcnghia:
@googlebot I signed it!

tcnghia:
For downstream I think we should symlink this.

cc @nak3 @ZhiminXiang

tcnghia:
@nak3 rebased, can you please take another look?

PR #64, opened 2020-08-06
Revert "Revert "Removes unwanted timeout setting (#22)" (#39)"

tcnghia:
This reverts commit c2d27c44d14b7b5259ab04de248b9f0e439f9bc9.

Since there are more and more users reporting that the timeout is affecting their gRPC usage, I am going to roll this forward again. This is potentially breaking change, however ingress implementers have been looped in on Slack and bugs (and the empty value was permitted in the API, albeit with unclear meaning).

I will follow up with adding some reasonable conformance testing for the meaning of "no timeout", as well as more documentations.

/assign @vagababov

### DIFF
--- a/pkg/apis/networking/v1alpha1/ingress_defaults.go
+++ b/pkg/apis/networking/v1alpha1/ingress_defaults.go
@@ -18,12 +18,8 @@ package v1alpha1
 
 import (
 	"context"
-	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
-
-	"knative.dev/networking/pkg/apis/config"
 )
 
 // SetDefaults populates default values in Ingress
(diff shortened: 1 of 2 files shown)


tcnghia:
cc @richterdavid @mattmoor @jmprusi (`net-kourier`) @ZhiminXiang (`net-istio`) @concaf @ilackarms @hbagdi

mattmoor:
This isn't a *potentially* breaking change, it is a breaking change.  It broke Contour and it breaks Gloo.


> and the empty value was permitted in the API, albeit with unclear meaning

The meaning was not unclear, it was defaulted, and we now expect a very specific behavior when it isn't specified which has no test coverage in kingress conformance.

Honestly, I don't think that we should move forward this aggressively with such a change without an owner from every implementation signing off on the breaking change, or being known to work.  I think this gets you Istio, Contour, and Kourier.  If Gloo, Ambassador, and Kong are OK with us potentially breaking their integration, then I'll shut up, but I am pretty uncomfortable with us rolling this forward KNOWING that it breaks at least one of our integrations still.

---

Given that we (Knative) are the exclusive (supported) client of this API perhaps a better way to roll out this change would be to:
1. Change the route controller to *specify* this value (so defaulting doesn't matter because we don't rely on it).
2. Specify the semantics and add a conformance test.
3. Given suitable bake time on `2.` revert `1.`.

I think `1.` and `2.` are doable in ~a day.

tcnghia:
+1 to setting a value in Route controller. I staged https://github.com/knative/serving/pull/8965  .

After that is in we can add conformance tests and when things are implemented properly downstream we will revert https://github.com/knative/serving/pull/8965 .

PR #95, opened 2020-08-15
Prefer t.Cleanup over test.CleanupOnInterrupt and defer cancel.

mattmoor:
This mostly shifts the kingress conformance tests to use t.Cleanup.  We still
return `context.CancelFunc` to cleanup resources so that tests like TestUpdate
can choose to eagerly clean up resources before the entire test is complete.

### DIFF
--- a/test/conformance/ingress/util.go
+++ b/test/conformance/ingress/util.go
@@ -533,7 +533,7 @@ func CreateGRPCService(t *testing.T, clients *test.Clients, suffix string) (stri
 func createService(t *testing.T, clients *test.Clients, svc *corev1.Service) context.CancelFunc {
 	t.Helper()
 
-	test.CleanupOnInterrupt(func() {
+	t.Cleanup(func() {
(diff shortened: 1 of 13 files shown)


test/conformance/ingress/util.go
```diff
@@ -627,7 +622,7 @@ func CreateIngress(t *testing.T, clients *test.Clients, spec v1alpha1.IngressSpe
 		},
 		Spec: spec,
 	}
-	test.CleanupOnInterrupt(func() { clients.NetworkingClient.Ingresses.Delete(ing.Name, &metav1.DeleteOptions{}) })
+	t.Cleanup(func() { clients.NetworkingClient.Ingresses.Delete(ing.Name, &metav1.DeleteOptions{}) })
```
ZhiminXiang, line 625: Do we want to use EnsureCleanUp here (https://github.com/knative/networking/blob/4b262113649d4621e23c7267cab998e75d3cba6a/test/cleanup.go#L64) to make sure resources are cleaned up either on interrupt or when tests complete?

PR #102, opened 2020-08-20
Enable kingress composition.

mattmoor:
This adjusts the shared probing logic such that kingress resources will properly compose to implement our dataplane contract.  Today the services referenced from kingress are expected to respond to `K-Network-Probe: probe` and `K-Network-Hash: deadbeef` with the network hash they received.  However, the problem we had with composition was that this header was injected unconditionally, so the kingress itself couldn't implement the dataplane contract.

Our new logic leverages header matching to overwrite the network hash only when `K-Network-Hash: override` is specified, which enables us to probe the kingress directly, but avoid overriding values for this, which might come from probing an outer layer.

This makes the implementation of our dataplane contract a simple induction, where the base case is implemented by the service referenced by the kingress and the inductive step is implemented by the conditional header injection.

Fixes https://github.com/knative/serving/issues/9074

As part of this I also made our tests not hand indefinitely as a failure mode, but timeout after some modest duration.

/hold

I'm placing a hold on this, as it will likely require updated to downstream unit tests (I know net-contour needed some).  I'll remove the hold once I've gotten some eyeballs on the staged PRs, but this is RFAL now.

### DIFF
--- a/pkg/ingress/ingress.go
+++ b/pkg/ingress/ingress.go
@@ -52,12 +52,20 @@ func InsertProbe(ing *v1alpha1.Ingress) (string, error) {
 		if rule.HTTP == nil {
 			return "", fmt.Errorf("rule is missing HTTP block: %+v", rule)
 		}
+		probePaths := make([]v1alpha1.HTTPIngressPath, 0, len(rule.HTTP.Paths))
 		for i := range rule.HTTP.Paths {
-			if rule.HTTP.Paths[i].AppendHeaders == nil {
-				rule.HTTP.Paths[i].AppendHeaders = make(map[string]string, 1)
+			elt := rule.HTTP.Paths[i].DeepCopy()
+			if elt.AppendHeaders == nil {
+				elt.AppendHeaders = make(map[string]string, 1)
 			}
-			rule.HTTP.Paths[i].AppendHeaders[net.HashHeaderName] = hash
+			if elt.Headers == nil {
+				elt.Headers = make(map[string]v1alpha1.HeaderMatch, 1)
+			}
+			elt.Headers[net.HashHeaderName] = v1alpha1.HeaderMatch{Exact: net.HashHeaderValue}
+			elt.AppendHeaders[net.HashHeaderName] = hash
+			probePaths = append(probePaths, *elt)
 		}
+		rule.HTTP.Paths = append(probePaths, rule.HTTP.Paths...)
 	}
 
 	return hash, nil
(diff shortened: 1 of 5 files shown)


pkg/ingress/ingress.go
```diff
@@ -52,12 +52,20 @@ func InsertProbe(ing *v1alpha1.Ingress) (string, error) {
 		if rule.HTTP == nil {
 			return "", fmt.Errorf("rule is missing HTTP block: %+v", rule)
 		}
+		probePaths := make([]v1alpha1.HTTPIngressPath, 0, len(rule.HTTP.Paths))
 		for i := range rule.HTTP.Paths {
```
julz, line 56: kind of surprised nothing needed to change in ingress_test.go for this.. is it missing a test?

julz (commented):
code & general idea lgtm - seems like the tests in ingress_test.go don't really cover the side-effect ingress modifications it does though, which are now non-trivial, worth adding some coverage?

pkg/ingress/ingress.go
mattmoor: Yeah, I'll check it out this morning.  I was mainly focused on fixing unit tests and whether it blew up any of the kingress using it.  I think currently net-contour is the only implementation that doesn't "just" auto-import this, which is sort of cool.

pkg/ingress/ingress_test.go
```diff
@@ -131,17 +131,40 @@ func TestInsertProbe(t *testing.T) {
...
-			after := len(test.ingress.Spec.Rules[0].HTTP.Paths[0].AppendHeaders)
-			if before+1 != after {
-				t.Errorf("InsertProbe() left %d headers, wanted %d", after, before+1)
+
+			afterPaths := len(test.ingress.Spec.Rules[0].HTTP.Paths)
+			if beforePaths+beforePaths != afterPaths {
+				t.Errorf("InsertProbe() %d paths, wanted %d", afterPaths, beforePaths+beforePaths)
+			}
+
+			// Check the matches at the begining.
```
mattmoor, line 150: Found potential misspelling(s):
```suggestion
			// Check the matches at the beginning.
```

pkg/ingress/ingress.go
mattmoor: ok, the test is written in a funny way.  I expanded to cover this.

mattmoor:
/unhold

The downstream PRs are all green.

tcnghia:
Thanks for this brilliant and practical change. Can't wait to leverage this to make probing work with `RewriteHost` with iterations dialed to 15.

/lgtm

PR #107, opened 2020-08-21
Rework RewriteHost to require splits.

mattmoor:
This reworks the RewriteHost semantics to require splits and specify the backing service.  The conformance test fronts one kingress with another and an ExternalName service with the private load balanced in between.

This is precisely how Serving exposes the cluster-local services (with the exception of port, which I plan to PR!), and enables RewriteHost to happen as a single layer, multiple layers, and to properly probe!

/hold

I still need to update the `Validate()` logic to reflect the new structure, but this achieves RewriteHost via the composition enabled with the new probing logic.  The composition stuff also potentially sets the stage for some REALLY cool things we can do with header/path/method(todo)-based routing, which I find really exciting 😍 

My remaining hitlist before this is good-to-go:
1. Patch net-contour's endpoint-probing to also rewrite hosts.
2. Patch Serving's placeholder service to specify ports
3. Update this PR with Validate() changes.
4. Test the waters with net-istio.

cc @tcnghia @julz


test/conformance/ingress/util.go
```diff
@@ -551,6 +538,28 @@ func createService(t *testing.T, clients *test.Clients, svc *corev1.Service) con
 	}
 }
 
+func createExternalNameService(t *testing.T, clients *test.Clients, target, gatewayDomain string) context.CancelFunc {
+	targetName := strings.Split(target, ".")
```
vagababov, line 542: splitn? But in general we construct target only to deconstruct it here. May be just pass the 2 parts?

mattmoor: done

julz (commented):
I like this direction a _lot_.

/lgtm

PR #121, opened 2020-08-26
Remove ingressspec.visibility

MIBc:
* remove visibility in IngressSpec

As discussed in https://github.com/knative/serving/pull/8984#discussion_r469878265. The visibility in IngressSpec is useless.

### DIFF
--- a/pkg/apis/networking/v1alpha1/ingress_types.go
+++ b/pkg/apis/networking/v1alpha1/ingress_types.go
@@ -110,9 +110,6 @@ type IngressSpec struct {
 	// A list of host rules used to configure the Ingress.
 	// +optional
 	Rules []IngressRule `json:"rules,omitempty"`
-
-	// Visibility setting.
-	Visibility IngressVisibility `json:"visibility,omitempty"`
 }
 
 // IngressVisibility describes whether the Ingress should be exposed to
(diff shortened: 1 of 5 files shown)


pkg/apis/networking/v1alpha1/ingress_types.go
```diff
@@ -110,9 +110,6 @@ type IngressSpec struct {
 	// A list of host rules used to configure the Ingress.
 	// +optional
 	Rules []IngressRule `json:"rules,omitempty"`
-
-	// Visibility setting.
-	Visibility IngressVisibility `json:"visibility,omitempty"`
```
nak3, line 115: @tcnghia  Please allow me double check. We can remove it now ? Or we should add "obsolete" comment and remove it after 9 months?  

I am worried as some KIngress still refer to this field, so they need to update as soon as this change was pulled.

e.g. contour
https://github.com/knative-sandbox/net-contour/blob/31f50d1c2c86d19245d42b34a9532a833daf990d/pkg/reconciler/contour/resources/kingress.go#L46

MIBc: Sincerely, we can keep my change https://github.com/knative/serving/pull/8984 which only involve serving. I suggest don't remove  spec.Visibility temporary.

nak3:
https://github.com/knative/networking/pull/129 made ingress.spec.visibility deprecated and ingress.spec.visibility will always nil now.

PR #129, opened 2020-08-31
Make ingress.spec.visibility deprecated

nak3:
`ingress.spec.visibility` was used for a fallback when
`spec.rules.visibility` isn't set. But it is not used anymore
actually as route controller always set it.

https://github.com/knative/serving/blob/4c2ae6597b07c8ac05a9731e647d60164940a95e/pkg/reconciler/route/resources/ingress.go#L223

Hence this patch changes to:
- make `ingress.spec.visibility` deprecated.
- drop `+optional` from `spec.rules.visibility`.
- set default to `spec.rules.visibility` for sure.

Co-authored-by: Lvjiawei <lvjiawei@cmss.chinamobile.com>

/cc @tcnghia @mattmoor @ZhiminXiang


nak3:
(Rebased)

Issue #136, opened 2020-09-03
Deprecate ingress.status.LoadBalancer


nak3:
There are three LB status in `ingress.status`, LoadBalancer, PublicLoadBalancer and PrivateLoadBalancer.

https://github.com/knative/networking/blob/f8dacf70c1e8a7daa7dd8847f42f3eb70fc5496a/pkg/apis/networking/v1alpha1/ingress_types.go#L304-L315

`PublicLoadBalancer` and `PrivateLoadBalancer` make sense but `LoadBalancer` does not in the status field.

#### net-istio and net-kourier
- external LB or internal LB is assigned in the LoadBalancer based on `ingress.Spec.Visibility`. But `ingress.Spec.Visibility` was deprecated.

#### net-contour
- net-contour does not use it and always set LoadBalancer to nil.

PR #137, opened 2020-09-04
Deprecate ingress.status.LoadBalancer status field

nak3:
This patch deprecates ingress.status.LoadBalancer. Use
`PublicLoadBalancer` and `PrivateLoadBalancer` instead.

Fixes https://github.com/knative/networking/issues/136


tcnghia (approved):
/lgtm

Issue #113, opened 2020-08-22
Require kingress to implement the dataplane contract


mattmoor:
We should add a kingress conformance test that validates conformance to our dataplane contract.

### Base Case
If we send:
```
K-Network-Probe: probe
K-Network-Hash: override
```
Then we should get back a response with `K-Network-Hash: {hash reflecting config}`.

### Inductive Step
If we send:
```
K-Network-Probe: probe
K-Network-Hash: {hex}
```
Then we should get back a response with `K-Network-Hash: {hex}`.

---

This verifies that we can enable composition of kingress using ExternalName services.

PR #138, opened 2020-09-06
Add KIngress conformance test for probe request

nak3:
This patch adds a kingress conformance test that validates conformance
to dataplane contract.

If probe request has a header:

```
K-Network-Hash: override
```

Then we should get back a response with `K-Network-Hash: {hash
reflecting config}`.

If probe request has a header

```
K-Network-Hash: {hex}
```

Then we should get back a response with `K-Network-Hash: {hex}`.

Fixes https://github.com/knative/networking/issues/113

### DIFF
--- a/test/conformance/ingress/headers.go
+++ b/test/conformance/ingress/headers.go
@@ -17,6 +17,7 @@ limitations under the License.
 package ingress
 
 import (
+	"fmt"
 	"net/http"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


nak3:
/hold

I think that the test still fails until https://github.com/knative-sandbox/net-istio/pull/263 is merged, because a deprecated field is still valid in net-istio's `ingress.Spec`.

test/conformance/ingress/run.go
```diff
@@ -32,6 +32,7 @@ func RunConformance(t *testing.T) {
 	t.Run("grpc", TestGRPC)
 	t.Run("grpc/split", TestGRPCSplit)
 
+	t.Run("headers/probe", TestProbeHeaders)
```
mattmoor, line 35: I think we need to add this under alpha or beta and socialize the new addition a bit.

test/conformance/ingress/headers.go
```diff
@@ -26,10 +27,77 @@ import (
...
+							ServiceNamespace: test.ServingNamespace,
+							ServicePort:      intstr.FromInt(port),
+						},
+					}},
+				}},
+			},
+		}},
+	})
+
+	bytes, _ := ingress.ComputeHash(ing)
```
mattmoor, line 61: may as well check the error and `t.Error` if it is non-nil.

test/conformance/ingress/run.go
nak3: Alright. I move it to beta as it is not a brand-new feature but we just need to socialize it.

test/conformance/ingress/headers.go
nak3: Fixed. Thank you!

PR #139, opened 2020-09-06
Make kingress conformance amenable to KinD

mattmoor:
Two main changes:
1. Use `imagePullPolicy: IfNotPresent` to support preloading the test images with `ko` + `kind.local`
2. Don't require the service exposed in the kingress status to have type LoadBalancer when we've been handed an `--ingressendpoint`.  This is indicative of local environments like KinD or minikube where the `type: LoadBalancer` service may have been changes to `type: NodePort`.

### DIFF
--- a/test/conformance/ingress/util.go
+++ b/test/conformance/ingress/util.go
@@ -874,7 +874,7 @@ func CreateDialContext(t *testing.T, ing *v1alpha1.Ingress, clients *test.Client
 				return nil, err
 			}
 			for _, sp := range svc.Spec.Ports {
-				if fmt.Sprintf("%d", sp.Port) == port {
+				if fmt.Sprint(sp.Port) == port {
 					return dial(ctx, "tcp", fmt.Sprintf("%s:%d", pkgTest.Flags.IngressEndpoint, sp.NodePort))
 				}
 			}
--- a/test/conformance/ingress/util.go
+++ b/test/conformance/ingress/util.go
@@ -106,8 +106,9 @@ func CreateRuntimeService(t *testing.T, clients *test.Clients, portName string) 
 		},
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{{
-				Name:  "foo",
-				Image: pkgTest.ImagePath("runtime"),
+				Name:            "foo",
+				Image:           pkgTest.ImagePath("runtime"),
+				ImagePullPolicy: corev1.PullIfNotPresent,
 				Ports: []corev1.ContainerPort{{
 					Name:          portName,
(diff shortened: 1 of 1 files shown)


mattmoor:
cc @nak3 @ZhiminXiang

mattmoor:
I feel like `--ingressendpoint` is going to be inadequate because we won't hit the same port for both http and https.

Perhaps the proper fix would be to allow the service to be NodePort when an IP is specified via `--ingressendpoint` and select the appropriate node port from the service...

test/conformance/ingress/util.go
```diff
@@ -854,28 +861,43 @@ func CreateDialContext(t *testing.T, ing *v1alpha1.Ingress, clients *test.Client
...
+		// If we're using a manual --ingressendpoint then don't require
+		// "type: LoadBalancer", which may not play nice with KinD
+		return func(ctx context.Context, _ string, address string) (net.Conn, error) {
+			_, port, err := net.SplitHostPort(address)
+			if err != nil {
+				return nil, err
+			}
+			for _, sp := range svc.Spec.Ports {
+				if fmt.Sprintf("%d", sp.Port) == port {
+					return dial(ctx, "tcp", fmt.Sprintf("%s:%d", pkgTest.Flags.IngressEndpoint, sp.NodePort))
```
vagababov, line 878: you already have string representation `port` so just `pkgTest.Flags.IngressEndpoint + ":" + port`?

vagababov (commented):
The better name for the branch would be `kinder` 😬

test/conformance/ingress/util.go
mattmoor: NodePort is different

test/conformance/ingress/util.go
```diff
@@ -854,28 +861,43 @@ func CreateDialContext(t *testing.T, ing *v1alpha1.Ingress, clients *test.Client
...
+
+		// If we're using a manual --ingressendpoint then don't require
+		// "type: LoadBalancer", which may not play nice with KinD
+		return func(ctx context.Context, _ string, address string) (net.Conn, error) {
+			_, port, err := net.SplitHostPort(address)
+			if err != nil {
+				return nil, err
+			}
+			for _, sp := range svc.Spec.Ports {
+				if fmt.Sprintf("%d", sp.Port) == port {
```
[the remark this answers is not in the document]
mattmoor: done

mattmoor:
Here's a run of kingress conformance passing with this vendored: https://github.com/mattmoor/net-contour/runs/1079033774?check_suite_focus=true

tcnghia (approved):
/lgtm

PR #144, opened 2020-09-09
Make sure update-codegen also updates the config network CM

vagababov:
- drink coffee
- pull new deps
- update the script
- update tools.go
- add some example periods to verify the generation works

/assign mattmoor @tcnghia


config/config-network.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: "3317ccca"
+    knative.dev/example-checksum: "2653766a"
```
mattmoor, line 23: Just to check: if this isn't updated, will the tests fail?

mattmoor (commented):
/lgtm
/approve
/hold

Hold for the question

config/config-network.yaml
tcnghia: Didn't fail in my PR.

vagababov: Dunno, do we call verify codegen?

vagababov:
In general it should, since I've added this to the `update-codegen` so it'll generate a diff now.
/unhold

PR #153, opened 2020-09-11
Add a dummy file so we can import configs

nak3:
This is same with caching's https://github.com/knative/caching/commit/64d6830ca6ddf965bf2ac7354526f502451bf138 but for networking.


mattmoor (commented):
/lgtm

PR #154, opened 2020-09-11
Move Kcert and KIngress CRDs from serving to networking repo

nak3:
This patch moves Kcert and KIngress CRDs from serving to networking repo.


mattmoor (commented):
/lgtm

PR #156, opened 2020-09-11
Remove tagHeaderBasedRouting example from config-network.yaml

nak3:
This patch removes `tagHeaderBasedRouting` example from config-network.yaml.
It was removed by https://github.com/knative/serving/commit/b564cda38a6468703e85c3f544eacb12baecd1fc.

Also this patch tweaks a filepath against configmap in update-checksums.sh.

/cc @tcnghia @vagababov


tcnghia (approved):
/lgtm

Issue #91, opened 2020-08-15
conformance test for omitted timeout


mattmoor:
cc @tcnghia @richterdavid 

/kind good-first-issue
/kind cleanup

mattmoor:
We might also consider deprecating this field and removing the other tests, which occasionally flake for me.

mattmoor:
cc @nak3 

Want to knock these off too? 😇

nak3:
Sure, I take it :smile_cat:

PR #157, opened 2020-09-13
Add conformance test for omitted timeout

nak3:
Currently timeout sets to super long time until Ingress providers
implements the "no timeout".
However there are no conformance test for it so we cannot see if they
implemented it or not.

This patch adds the conformance test for it.

Fixes https://github.com/knative/networking/issues/91


tcnghia (approved):
/lgtm

PR #167, opened 2020-09-15
Some followup to the #164

vagababov:
Since I didn't have time to comment :/
I think 64 is mostly theoretical here, since NS can't be longer than 63 anyway, but with some custom stuff it might?
/assign @shreejad @ZhiminXiang

### DIFF
--- a/pkg/ingress/ingress.go
+++ b/pkg/ingress/ingress.go
@@ -20,7 +20,6 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -100,10 +99,9 @@ func ExpandedHosts(hosts sets.String) sets.String {
 	expanded := make(sets.String, len(hosts)*len(allowedSuffixes))
 	for _, h := range hosts.List() {
 		for _, suffix := range allowedSuffixes {
-			if strings.HasSuffix(h, suffix) {
-				trimmedHost := strings.TrimSuffix(h, suffix)
-				if isValidTopLevelDomain(trimmedHost) {
-					expanded.Insert(trimmedHost)
+			if th := strings.TrimSuffix(h, suffix); suffix == "" || len(th) < len(h) {
+				if isValidTopLevelDomain(th) {
+					expanded.Insert(th)
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/ingress/ingress.go
```diff
@@ -100,22 +99,33 @@ func ExpandedHosts(hosts sets.String) sets.String {
...
-	topLevelDomain := parts[len(parts)-1]
-	_, err := strconv.Atoi(topLevelDomain)
-	return (err != nil)
+	tld := parts[len(parts)-1]
+	if len(tld) > 63 {
+		return false
+	}
+	for _, c := range []byte(tld) {
+		c -= '0'
+		if c > 9 {
```
shreejad, line 125: Do all allowed characters in a domain have a byte value larger than that of '0'?

vagababov: hmmm, interesting, `-` I think before numbers. let's do a more precise check.

shreejad:
Sorry @vagababov, did not know the PR would automatically close when it got an approval. Is that a setting I needed to disable somewhere?

vagababov:
> Sorry @vagababov, did not know the PR would automatically close when it got an approval. Is that a setting I needed to disable somewhere?

Not a problem, you don't have to wait for me to review all the PRs — I'd hate to be such a blocker :-)
I just didn't get to comment, but that's alright.

pkg/ingress/ingress.go
vagababov: So, `-` is smaller than `0`, so I fixed that. Added a test.
Also improved how we generate test failures.

vagababov:
PTAL

pkg/ingress/ingress.go
```diff
@@ -100,22 +99,32 @@ func ExpandedHosts(hosts sets.String) sets.String {
...
 	parts := strings.Split(domain, ".")
-	topLevelDomain := parts[len(parts)-1]
-	_, err := strconv.Atoi(topLevelDomain)
-	return (err != nil)
+	tld := parts[len(parts)-1]
+	if len(tld) > 63 {
+		return false
+	}
+	for _, c := range []byte(tld) {
+		if c == '-' || c > '9' {
```
shreejad, line 124: Nit: c > 9

shreejad: My bad, ignore this

vagababov: 👍

shreejad:
I think this is where the error was being thrown by Istio where the TLD is declared invalid - https://github.com/istio/istio/blob/eb63aae13235c98f99cea6591d4401c569576949/pkg/config/validation/validation.go#L170
They only seem to be checking whether the TLD is all-numeric. Does it make sense to open an Istio bug for the 64 character limit that you've added here @vagababov ? I'm not sure what that edge case would result in right now

vagababov:
This check `labels.IsDNS1123Label(label)` does that.

shreejad:
I see, I thought this limit was specifically for TLDs. Why are we particularly checking the length of only the TLD if the length limit applies to all parts? Wouldn't a TLD having length > 63 mean that the other trimmed short-names and the original domain name is also invalid?
In [ingress_test.go L62](https://github.com/vagababov/networking/blob/13dcea8824bf9eb826d5597afbc2f0ac601783cd/pkg/ingress/ingress_test.go#L62), "service."+strings.Repeat("s", 64)+".svc" would be invalid?

vagababov:
No, it splits and validates each element to be a valid 1123 label.

shreejad:
I had assumed (naively I guess) that the entire FQDN verification would have already occurred at some prior point, and this code was just to check that given a valid FQDN, we don't create an invalid trimmed short-name

shreejad:
> No, it splits and validates each element to be a valid 1123 label.

IIUC, for a long namespace (len > 64), ExpandedHosts() will return "service.<long_namespace>.svc" and "service.<long_namespace>.svc.cluster.local", and would have avoided adding "service.<long_namespace>". Are the first 2 domain names that we return valid?

vagababov:
The whole name can be long. Each component shouldn't. Or I am misunderstanding.

PR #169, opened 2020-09-16
[release-0.17] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign tcnghia
/cc tcnghia


tcnghia:
/hold

I am not sure if we want to cause a release just because of a vendor change in `test-infra` 
cc @nak3 @ZhiminXiang

ZhiminXiang:
If this is the only for `test-infra` update, I think we should close this one.

PR #178, opened 2020-09-21
Various fixes around the networking repo

vagababov:
- use transport clone to get default (thanks @julz)
- make methoods that are not used in other repos private
- other small fixes

/assign @tcnghia mattmoor

### DIFF
--- a/test/conformance/ingress/class.go
+++ b/test/conformance/ingress/class.go
@@ -71,7 +71,9 @@ func verifyIngressWithAnnotations(ctx context.Context, t *testing.T, clients *te
 	annotations map[string]string, backend *v1alpha1.IngressBackend) {
 	t.Helper()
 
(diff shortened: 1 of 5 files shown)


test/conformance/ingress/class.go
```diff
@@ -86,6 +87,7 @@ func verifyIngressWithAnnotations(ctx context.Context, t *testing.T, clients *te
 		},
 		OverrideIngressAnnotation(annotations),
 	)
+	t.Cleanup(cancel)
```
mattmoor, line 90: It does this internally already and optionally returns it in case folks want to clean it up earlier (the way TestUpdate does).

vagababov: 😠 
Done.

mattmoor (commented):
/lgtm

PR #184, opened 2020-09-23
Support skip-tests option to skip some specific tests in alpha or beta features

nak3:
Currently Alpha or Beta conformance tests must be implemented when we enabled the flag.
This patch allows users to skip one or some specific tests in alpha or beta features.

Fixes https://github.com/knative/networking/issues/183

PR #187, opened 2020-09-24
Feature-flag gated tests to validate auto http2/gRPC detection with empty port

rafaeltello:
Creates new "emptyport" tests to validate protocol used when no port is specified. Currently only 1 of the tests runs all the time, the others are gated behind a new feature flag.

We originally intended to add a more complex support matrix in this change. However, while we can hack the `http.Transport.TLSNextProto` to force a http 1.1 connection, the way the dialer is implemented (by using a TCP connection to the ingress endpoint) makes it meaningless in this test. Conversely, using `http2.Transport` to force a http2-only connection requires us to create an equivalent version of https://github.com/knative/networking/blob/master/test/conformance/ingress/util.go#L902 using a slightly different signature.

We still want to do this, but will probably occur in parallel with the rest of development.

Work is being tracked by https://github.com/knative/serving/issues/4283

### DIFF
--- a/test/conformance/ingress/emptyport.go
+++ b/test/conformance/ingress/emptyport.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Knative Authors
+Copyright 2020 The Knative Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/test/conformance/ingress/emptyport.go
+++ b/test/conformance/ingress/emptyport.go
@@ -95,8 +95,7 @@ func TestHTTP2AndEmptyPort(t *testing.T) {
 	}
 
 	if want, got := 2, ri.Request.ProtoMajor; want != got {
-		// DO NOT SUBMIT
-		t.Fatalf("ProtoMajor = %d, wanted %d", got, want)
+		t.Errorf("ProtoMajor = %d, wanted %d", got, want)
 	}
 }
 
--- a/test/conformance/ingress/emptyport.go
+++ b/test/conformance/ingress/emptyport.go
@@ -100,7 +100,6 @@ func TestHTTP2AndEmptyPort(t *testing.T) {
 }
 
 // TestGRPCWithEmptyPort verifies that a nameless port can establish a basic GRPC connection.
-// DO NOT SUBMIT: This breaks.
 func TestGRPCWithEmptyPort(t *testing.T) {
 	t.Parallel()
 	defer logstream.Start(t)()
--- a/test/conformance/ingress/emptyport.go
+++ b/test/conformance/ingress/emptyport.go
@@ -1,0 +1,163 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ingress
+
+import (
+	"context"
+	"net"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"knative.dev/networking/pkg/apis/networking/v1alpha1"
+	"knative.dev/networking/test"
+	ping "knative.dev/networking/test/test_images/grpc-ping/proto"
+	"knative.dev/pkg/test/logstream"
+)
+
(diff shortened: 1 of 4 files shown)


igorbelianski (approved):

test/conformance/ingress/emptyport.go
```diff
@@ -0,0 +1,156 @@
+/*
+Copyright 2019 The Knative Authors
```
chaodaiG, line 2:
```suggestion
Copyright 2020 The Knative Authors
```

test/conformance/ingress/emptyport.go
```diff
@@ -0,0 +1,156 @@
...
+							ServicePort:      intstr.FromInt(port),
+						},
+					}},
+				}},
+			},
+		}},
+	})
+
+	ri := RuntimeRequest(ctx, t, client, "http://"+name+".example.com")
+	if ri == nil {
```
chaodaiG, line 93: This test can succeed if ri is nil or some other condition, is it expected?

tcnghia:
@rafaeltello can https://github.com/knative/networking/pull/184 be useful for this purpose? can you please check with @nak3? thanks!

test/conformance/ingress/emptyport.go
rafaeltello: This is essentially https://github.com/knative/networking/blob/master/test/conformance/ingress/basic.go#L60 with an empty port, so we intentionally want to keep the test as close as possible as that one.

rafaeltello:
> @rafaeltello can #184 be useful for this purpose? can you please check with @nak3? thanks!

Yes!!! This is probably super useful, and I now need to merge with this file. Let me change that.

rafaeltello:
> @rafaeltello can #184 be useful for this purpose? can you please check with @nak3? thanks!

I ended up keeping my original flag, as the feature is not even Alpha yet. I did add a TODO to move this into alphaTests once the feature is implemented :)

test/conformance/ingress/run.go
```diff
@@ -93,6 +93,14 @@ func RunConformance(t *testing.T) {
 			t.Run(name, test)
 		}
 	}
+
+	// TODO(rafaeltc): Once the feature is partially implemented,
+	// move to alphaTests and remove flag.
```
tcnghia, line 98: Link to issue instead? Thanks

tcnghia (changes requested):
/assign @nak3

test/conformance/ingress/emptyport.go
```diff
@@ -0,0 +1,156 @@
...
+	"google.golang.org/grpc"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"knative.dev/networking/pkg/apis/networking/v1alpha1"
+	"knative.dev/networking/test"
+	ping "knative.dev/networking/test/test_images/grpc-ping/proto"
+	"knative.dev/pkg/test/logstream"
+)
+
+// TestHTTP1AndEmptyPort verifies that an empty port name uses HTTP1. This should be the current behavior.
+func TestHTTP1AndEmptyPort(t *testing.T) {
```
nak3, line 34: Can we add `t.Parallel()`?

test/conformance/ingress/emptyport.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	if want, got := 1, ri.Request.ProtoMajor; want != got {
+		t.Errorf("ProtoMajor = %d, wanted %d", got, want)
+	}
+}
+
+// TestHTTP2AndEmptyPort verifies that an empty port name uses HTTP2.
+// This is not the current behavior.
+func TestHTTP2AndEmptyPort(t *testing.T) {
```
nak3, line 69: Same with above. Can we add `t.Parallel()`?

nak3 (commented):
In addition these tests, should we add a traffic split test which has empty name port and non-empty name port mixed?

test/conformance/ingress/emptyport.go
```diff
@@ -0,0 +1,156 @@
...
+			},
+		}},
+	})
+
+	ri := RuntimeRequest(ctx, t, client, "http://"+name+".example.com")
+	if ri == nil {
+		return
+	}
+
+	if want, got := 2, ri.Request.ProtoMajor; want != got {
```
nak3, line 97: I'm sorry I may be stupid. But `TestHTTP1AndEmptyPort` and `TestHTTP2AndEmptyPort` are same tests and using same `runtime` image. How this test can get protocol version 2?

test/conformance/ingress/emptyport.go
rafaeltello: Will do! Sorry, I created these tests before that change went in, and missed copying it.

test/conformance/ingress/emptyport.go
rafaeltello: Oh! No, I'm the stupid one :( you're absolutely right; I had done some additional work here, but then I reverted most of it and didn't notice the two test cases were the same.
Let me rewrite this today.

rafaeltello:
Thanks to @nak3 comment we discovered a more fundamental testability issue with this change. We need to test this in serving, not in networking, to actually exercise knative's eventual ability to autodetect http2/gRPC. These are k8s services.

I am closing this request, and sending equivalent ones for serving.

PR #188, opened 2020-09-25
[Automated] Update actions

mattmoor:
Produced via:
```shell
# meta: knative/.github
# main: knative/networking
cp $(find "${GITHUB_WORKSPACE}/meta/workflow-templates" -type f -name '*.yaml') \
  "${GITHUB_WORKSPACE}/main/.github/workflows/"
```
/cc tcnghia nak3 ZhiminXiang
/assign tcnghia nak3 ZhiminXiang

### DIFF
--- a/.github/workflows/knative-verify.yaml
+++ b/.github/workflows/knative-verify.yaml
@@ -57,7 +57,7 @@ jobs:
       shell: bash
       run: |
         pushd ./src/knative.dev/${{ github.event.repository.name }}
-        ./hack/update-codegen.sh
+        [[ ! -f hack/update-codegen.sh ]] || ./hack/update-codegen.sh
         popd
 
     - name: Verify


.github/workflows/knative-verify.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
       shell: bash
       run: |
         pushd ./src/knative.dev/${{ github.event.repository.name }}
-        ./hack/update-codegen.sh
+        [[ ! -f hack/update-codegen.sh ]] || ./hack/update-codegen.sh
```
nak3, line 60: Hmm... I think producing an error if `hack/update-codegen.sh` was missed instead of skipping `./hack/update-codegen.sh`.

@mattmoor What is the reason of this change?

mattmoor: context: these are all simply mirrors of `knative-sandbox/.github`'s

`knative/docs` doesn't have `hack/update-codegen.sh`, and probably shouldn't.  We've worked around this some places (e.g. net-kourier) by making it a thin shim to `update-deps.sh`, but that sucks.  The knobots have had a check like this for a while.

nak3: Thank yo! That makes sense.

PR #190, opened 2020-09-29
This fixes logger and fmt strings

vagababov:
It's trickier here :)
logger does not pad with string, fmt does sometimes :/


tcnghia (approved):
/lgtm

PR #191, opened 2020-09-30
Enable extended linting and fix the respective issues

markusthoemmes:
As per title, align the linters with serving.


tcnghia (approved):
/lgtm

PR #192, opened 2020-09-30
Migrate constants that are only used in Serving

markusthoemmes:
The dual to https://github.com/knative/serving/pull/9636/

/assign @tcnghia


tcnghia (approved):
/lgtm

Issue #195, opened 2020-10-02
Drop deprecated generation field


dprotaso:
ie.

https://github.com/knative/networking/blob/52c85a370e452eb9217ee9531b7d80bbb4b23da3/pkg/apis/networking/v1alpha1/ingress_types.go#L91-L100

PR #199, opened 2020-10-02
Remove deprecated generation field

nak3:
This patch removes DeprecatedGeneration from ingress spec.
It was deprecated since https://github.com/knative/serving/pull/2923.

/cc @dprotaso @tcnghia @ZhiminXiang 

Fixes https://github.com/knative/networking/issues/195


tcnghia (approved):
/lgtm

PR #207, opened 2020-10-08
Improve test error debugging.

jpeach:
Add the quoted object name (fully qualified if possible) to test error
messages to make it easier for testers to understand why tests fail.

Signed-off-by: James Peach <jpeach@vmware.com>


mattmoor (approved):
/lgtm

Issue #208, opened 2020-10-08
Test downstream integrations


mattmoor:
See: https://github.com/knative/eventing/pull/4260

cc @nak3 @ZhiminXiang @tcnghia 

/kind good-first-issue

PR #213, opened 2020-10-09
Add downstream test with serving and net-*

nak3:
This patch adds downstream tests with serving, net-istio, net-contour, net-kourier, net-certmanager and net-http01.

Fixes https://github.com/knative/networking/issues/208

### DIFF
--- a/.github/workflows/knative-downstream.yaml
+++ b/.github/workflows/knative-downstream.yaml
@@ -183,3 +183,85 @@ jobs:
       with:
         upstream-module: knative.dev/${{ github.event.repository.name }}
         downstream-module: knative.dev/net-kourier
+
+  downstream-net-certmanager:
+    name: Net Certmanager
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


tcnghia (approved):
/lgtm
💯

.github/workflows/knative-downstream.yaml
```diff
@@ -0,0 +1,267 @@
...
+    branches: [ 'master', 'release-*' ]
+
+jobs:
+
+  downstream-serving:
+    name: Serving
+    strategy:
+      matrix:
+        go-version: [1.14.x]
+        platform: [ubuntu-latest]
```
mattmoor, line 28: Could we just use the downstream repo as a dimension of the matrix?  what varies job-to-job?

cc @n3wscott

nak3: Yes we can. Updated on https://github.com/knative/networking/pull/227. Thank you.

PR #227, opened 2020-10-17
Use the downstream repo as a dimension of the matrix

nak3:
This patch changes to use the downstream repo as a dimension of the
matrix.


tcnghia (approved):
/lgtm

PR #233, opened 2020-10-19
[wip] use reconciler-test as the base test framework

dprotaso:
Realizing https://github.com/knative/pkg/issues/1563

Downstream tests will fail since the flags are different

#### TODOs:
- [ ] add logstream support back
- [ ] Downstream / knative-sandbox-net-istio 
- [x] https://github.com/knative-sandbox/net-contour/pull/319
- [ ] Downstream / knative-sandbox-net-kourier
- [ ] Downstream / knative-sandbox-net-certmanager
- [ ] Downstream / knative-sandbox-net-http01


nak3:
@dprotaso Could you please update this? I will try to apply this to net-kourier and some other net-*.

dprotaso:
From this mornings [slack discussion](https://knative.slack.com/archives/CA9RHBGJX/p1610630625009700?thread_ts=1609868966.002100&cid=CA9RHBGJX):

> so it’s a bit confusing cause there are two frameworks in reconciler-test. The PR you linked uses the one I authored and is incremental to what networking and serving have but it doesn’t address a bunch of needs from eventing. So we’re trying to reconcile serving/networking needs into the larger framework that covers eventing’s use cases
>
> But we need input on
> 1) authoring tests
> 2) downstream consuming tests
> 3) anything else specific that say networking needs

dprotaso:
Going to close this - this version of reconciler test isn't being iterated on anymore and new work has progressed in a different iteration

PR #243, opened 2020-10-22
Use context from function args in WaitForIngressState

nak3:
This makes a tiny change that use the context from the function args

It seem that the context was added to the function args during k8s
1.18 libs update in 731bfc0, but we missed to use it.

/cc @dprotaso  @tcnghia @mattmoor


vagababov (commented):
/lgtm

tcnghia (approved):
/lgtm

markusthoemmes (approved):
/lgtm
/approve

PR #258, opened 2020-11-01
Move the constant to the networking package where it belongs

vagababov:
It should reside here probably.

/assign @tcnghia


tcnghia (approved):
/lgtm

PR #263, opened 2020-11-06
Enable tparallel linter and make sure all subtests are parallel

markusthoemmes:
Honestly not quite sure this is what we want. But this mostly looks like parallel requests anyway so :man_shrugging:.

/assign @nak3 @tcnghia


tcnghia (approved):
/lgtm

PR #265, opened 2020-11-10
Use a different hostname per variant.

mattmoor:
These variants are run in parallel and part the hostname to determine the external name service to create, so when things are run concurrently two variants may try to create the same service.

I don't know why we haven't seen this flake previously or why it appears in such a pronounced way with Contour 1.10, but this should hopefully address it.

/assign @tcnghia 
/hold

Holding for downstream validation, though it should be benign even if it doesn't fix the downstream issue I'm seeing.


tcnghia (approved):
/lgtm

mattmoor:
/unhold

Alright this is benign, and I won't be able to use it until we unpin anyways, so yolo.

PR #267, opened 2020-11-10
This drops the somewhat gratuitous use of t.Parallel in TLS tests.

mattmoor:
Downstream I'm seeing odd failures in net-contour where it takes ~32s for the kingress to become ready, but the `verify HTTPS` variant doesn't run until the certificate has expired 5 minutes later...

```
util.go:956: Error making GET request: Get "https://ingress-conformance-2-tls-dnewllzp.example.com": x509: certificate has expired or is not yet valid: current time 2020-11-10T19:41:26Z is after 2020-11-10T19:30:22Z
```

In the above test, we updated the status as ready at `19:25:46.726947295`, but didn't attempt the curl until `19:41:26` when the certificate expired at `19:30:22` (exactly 5 minutes after the test logs start).

I suspect that the problem here is actually that the execution of things with `t.Parallel` is highly unpredictable, so this removes the fairly gratuitous use of `t.Parallel` around what should be a simple `Get` request.

/assign @vagababov @tcnghia

### DIFF
--- a/test/conformance/ingress/tls.go
+++ b/test/conformance/ingress/tls.go
@@ -60,17 +60,11 @@ func TestIngressTLS(t *testing.T) {
 		}},
 	})
 
-	t.Run("verify HTTP", func(t *testing.T) {
-		t.Parallel()
+	// Check without TLS.
+	RuntimeRequest(ctx, t, client, "http://"+name+".example.com")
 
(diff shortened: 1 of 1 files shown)


mattmoor:
I see, we did this for the linter.  I'll make a slightly more dramatic change 🙃 

/hold

tcnghia (approved):
/lgtm

PR #269, opened 2020-11-12
Fix a few more occurrences of divisive language

markusthoemmes:
As per title.

/assign @nak3 @tcnghia


pkg/status/status_test.go
```diff
@@ -191,17 +191,17 @@ func TestProbeLifecycle(t *testing.T) {
 		}
 	}()
 
-	// Dummy handler returning HTTP 500 (it should never be called during probing)
-	dummyRequests := make(chan *http.Request)
-	dummyHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		dummyRequests <- r
+	// Test handler returning HTTP 500 (it should never be called during probing)
```
JRBANCEL, line 194: This might be well intentioned, but `test` is super vague.

vagababov: `failingHandler`

markusthoemmes: point taken, I'll revisit!

vagababov (commented):
/lgtm

PR #275, opened 2020-11-18
Surface the CreateIngress test helper

markusthoemmes:
As per title. This makes it simple to create Ingresses that doesn't necessarily have to become ready.

/assign @vagababov @tcnghia @nak3


tcnghia (approved):
/lgtm

PR #277, opened 2020-11-19
Add a conformance test to make sure multiple rules work

markusthoemmes:
We're already checking paths and splits but we're not making sure that rules with different hosts work as expected currently. Well... now we do!

/assign @tcnghia @nak3 @mattmoor


mattmoor (commented):
/lgtm

PR #288, opened 2020-12-02
Fix the probe package with prober constants and a free func

vagababov:
- the constants were wrong
- a const was marked as var
- add a free-floating function that seems to be the only current usage that we have of those functions.

/assign @tcnghia


tcnghia (approved):
/lgtm

PR #289, opened 2020-12-02
Move the drainer helper from pkg to the networking.

vagababov:
This is part of #1679.
Move the drainer and delete the probe handler.
This probe handler is unused in general, but its function
is used inside drainer, so move that.

/assign @tcnghia


tcnghia (approved):
/lgtm

PR #290, opened 2020-12-03
Move the prober library to the networking

vagababov:
Interestingly only test uses anything from pkg.

For #1679


tcnghia (approved):
/lgtm

PR #292, opened 2020-12-03
Remove the pkg references for prober

vagababov:
And use the local ones instead.

For: https://github.com/knative/pkg/issues/1679

/assign @tcnghia


tcnghia (approved):
/lgtm

PR #306, opened 2020-12-11
Add httpOption to KIngress spec

### LINKED ISSUE #297: [HTTP Option support] Add HTTPOption into Kingress spec
This is the first part of the httpOption support work. We need to add httpOption into Kingress spec.

Full design: https://docs.google.com/document/d/1yVOGo-j1hI7oCpuWVIi5iBJwX1wKbOzWaV0QNJxCbaY

arghya88:
Signed-off-by: Arghya Sadhu <arghya88@gmail.com>

# Changes

- :gift: Add httpoption to kingress spec

/kind enhancement

Fixes #297

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/apis/networking/v1alpha1/ingress_defaults.go
+++ b/pkg/apis/networking/v1alpha1/ingress_defaults.go
@@ -35,7 +35,9 @@ func (is *IngressSpec) SetDefaults(ctx context.Context) {
 	for i := range is.Rules {
 		is.Rules[i].SetDefaults(ctx)
 	}
-
+	if is.HTTPOption == "" {
+		is.HTTPOption = HTTPOptionEnabled
+	}
 	// Deprecated, do not use.
 	is.DeprecatedVisibility = ""
 }
--- a/pkg/apis/networking/v1alpha1/ingress_types.go
+++ b/pkg/apis/networking/v1alpha1/ingress_types.go
@@ -100,6 +100,11 @@ type IngressSpec struct {
 	// +optional
 	Rules []IngressRule `json:"rules,omitempty"`
 
+	// HTTPOption is the option of HTTP. It has the following three values:
+	// `HTTPOptionEnabled`, `HTTPOptionDisabled`, `HTTPOptionRedirected`
+	// If it's not specified then it defaults to `HTTPOptionEnabled`.
+	HTTPOption HTTPOption `json:"httpOption,omitempty"`
+
 	// DeprecatedVisibility was used for the fallback when spec.rules.visibility
 	// isn't set.
 	//
@@ -108,6 +113,18 @@ type IngressSpec struct {
 	// +optional
 	DeprecatedVisibility IngressVisibility `json:"visibility,omitempty"`
 }
+
+type HTTPOption string
+
+const (
+	// The knative ingress will be able to serve HTTP connections.
+	HTTPOptionEnabled HTTPOption = "Enabled"
+	// The knative ingress will reject HTTP traffic
+	HTTPOptionDisabled HTTPOption = "Disabled"
+	// The knative will return redirection HTTP status for the clients,
+	// asking the clients to redirect their requests to HTTPS.
+	HTTPOptionRedirected HTTPOption = "Redirected"
+)
 
 // IngressVisibility describes whether the Ingress should be exposed to
 // public gateways or not.
--- a/pkg/apis/networking/v1alpha1/ingress_validation.go
+++ b/pkg/apis/networking/v1alpha1/ingress_validation.go
@@ -50,6 +50,7 @@ func (is *IngressSpec) Validate(ctx context.Context) *apis.FieldError {
 	for idx, tls := range is.TLS {
 		all = all.Also(tls.Validate(ctx).ViaFieldIndex("tls", idx))
 	}
+	all = all.Also(is.HTTPOption.Validate(ctx))
 	return all
 }
 
@@ -163,6 +164,17 @@ func (t *IngressTLS) Validate(ctx context.Context) *apis.FieldError {
 	}
 	if t.SecretNamespace == "" {
 		all = all.Also(apis.ErrMissingField("secretNamespace"))
+	}
+	return all
+}
+
+// Validate inspects and validates HTTPOption object.
+func (t HTTPOption) Validate(ctx context.Context) (all *apis.FieldError) {
+	switch t {
+	case HTTPOptionEnabled, HTTPOptionDisabled, HTTPOptionRedirected:
+		break
+	default:
+		all = all.Also(apis.ErrInvalidValue(t, "httpOption"))
 	}
 	return all
 }
--- a/pkg/apis/networking/v1alpha1/ingress_defaults_test.go
+++ b/pkg/apis/networking/v1alpha1/ingress_defaults_test.go
@@ -33,7 +33,9 @@ func TestIngressDefaulting(t *testing.T) {
 		name: "empty",
 		in:   &Ingress{},
 		want: &Ingress{
-			Spec: IngressSpec{},
+			Spec: IngressSpec{
+				HTTPOption: HTTPOptionEnabled,
+			},
 		},
 	}, {
 		name: "split-timeout-and-visibility-defaulting",
@@ -78,6 +80,7 @@ func TestIngressDefaulting(t *testing.T) {
 						}},
 					},
 				}},
+				HTTPOption: HTTPOptionEnabled,
 			},
 		},
 	}, {
@@ -134,6 +137,7 @@ func TestIngressDefaulting(t *testing.T) {
 						}},
 					},
 				}},
+				HTTPOption: HTTPOptionEnabled,
 			},
 		},
 	}}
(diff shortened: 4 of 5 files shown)


pkg/apis/networking/v1alpha1/ingress_types.go
```diff
@@ -121,6 +128,13 @@ const (
 	// IngressVisibilityClusterLocal is used to denote that the Ingress
 	// should be only be exposed locally to the cluster.
 	IngressVisibilityClusterLocal IngressVisibility = "ClusterLocal"
+	// The knative ingress will be able to serve HTTP connections.
+	HTTPOptionEnabled HTTPOption = "Enabled"
+	// The knative ingress will reject HTTP traffic
+	HTTPOptionDisabled HTTPOption = "Disabled"
```
ZhiminXiang, line 134: Per the discussion in the [doc](https://docs.google.com/document/d/1yVOGo-j1hI7oCpuWVIi5iBJwX1wKbOzWaV0QNJxCbaY/edit?disco=AAAAKYgMWoc), we will not support Disabled (at least in the beginning) as it may break the HTTP01 challenge contract of managed TLS feature.

pkg/apis/networking/v1alpha1/ingress_types.go
```diff
@@ -100,6 +100,11 @@ type IngressSpec struct {
 	// +optional
 	Rules []IngressRule `json:"rules,omitempty"`
 
+	// HTTPOption is the option of HTTP. It has the following three values:
+	// `HTTPOptionEnabled`, `HTTPOptionDisabled`, `HTTPOptionRedirected`
```
ZhiminXiang, line 104: Same here. We don't need HTTPOptionDisabled.

ZhiminXiang (commented):
Thanks for the PR! @arghya88 

Generally looks good to me!

pkg/apis/networking/v1alpha1/ingress_types.go
```diff
@@ -121,6 +128,13 @@ const (
 	// IngressVisibilityClusterLocal is used to denote that the Ingress
 	// should be only be exposed locally to the cluster.
 	IngressVisibilityClusterLocal IngressVisibility = "ClusterLocal"
+	// The knative ingress will be able to serve HTTP connections.
+	HTTPOptionEnabled HTTPOption = "Enabled"
```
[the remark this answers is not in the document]
arghya88: Done

arghya88:
@ZhiminXiang are you okay with below as suggested by @vagababov ?

> We might want to keep the constant/validation, just not use it in the route controller when ingress is constructed?

pkg/apis/networking/v1alpha1/ingress_types.go
ZhiminXiang: Sounds good to me.

vagababov: We should just comment that right now disabled is not used.

pkg/apis/networking/v1alpha1/ingress_validation.go
```diff
@@ -166,3 +167,12 @@ func (t *IngressTLS) Validate(ctx context.Context) *apis.FieldError {
 	}
 	return all
 }
+
+// Validate inspects and validates HTTPOption object.
+func (t HTTPOption) Validate(ctx context.Context) *apis.FieldError {
+	var all *apis.FieldError
+	if t != HTTPOptionEnabled && t != HTTPOptionDisabled && t != HTTPOptionRedirected {
```
vagababov, line 174: it is more customary to use `switch` here.

arghya88: Not sure I understand how. Can you pls suggest a change?

vagababov, line 174:
```suggestion
	switch t {
	 case HTTPOptionEnabled, HTTPOptionDisabled, HTTPOptionRedirected:
	    break
	 default:
	    return error
	    }
```

arghya88: Done

pkg/apis/networking/v1alpha1/ingress_validation.go
```diff
@@ -166,3 +167,16 @@ func (t *IngressTLS) Validate(ctx context.Context) *apis.FieldError {
 	}
 	return all
 }
+
+// Validate inspects and validates HTTPOption object.
+func (t HTTPOption) Validate(ctx context.Context) *apis.FieldError {
+	var all *apis.FieldError
```
vagababov, line 173:
```suggestion
func (t HTTPOption) Validate(ctx context.Context) (all *apis.FieldError) {
```

vagababov (commented):
Just a nit.
/lgtm

pkg/apis/networking/v1alpha1/ingress_validation.go
arghya88: Done

tcnghia (approved):
/lgtm

pkg/apis/networking/v1alpha1/ingress_types.go
```diff
@@ -109,6 +114,18 @@ type IngressSpec struct {
 	DeprecatedVisibility IngressVisibility `json:"visibility,omitempty"`
 }
 
+type HTTPOption string
+
+const (
+	// The knative ingress will be able to serve HTTP connections.
+	HTTPOptionEnabled HTTPOption = "Enabled"
+	// The knative ingress will reject HTTP traffic
+	HTTPOptionDisabled HTTPOption = "Disabled"
```
mattmoor, line 123: @ZhiminXiang I think we talked about removing this and biasing towards removing the exposure from kingress gateways.  Part of this was motivated by not wanted to work out the interaction with HTTP01 TLS challenges IIRC.

pkg/apis/networking/v1alpha1/ingress_defaults.go
```diff
@@ -35,7 +35,9 @@ func (is *IngressSpec) SetDefaults(ctx context.Context) {
 	for i := range is.Rules {
 		is.Rules[i].SetDefaults(ctx)
 	}
-
+	if is.HTTPOption == "" {
+		is.HTTPOption = HTTPOptionEnabled
+	}
```
mattmoor, line 40: This may cause downgrade test failures 🤔

mattmoor: I suspect this change is making "upgrade" testing fail here: https://github.com/knative/serving/pull/10386

pkg/apis/networking/v1alpha1/ingress_types.go
ZhiminXiang: yes, we had the discussion in the design doc.

Here Victor and I thought we could still keep it in the IngressSpec but make Route controller not set "Disabled" option. With this way, the HTTP01 challenge will not be interacted within Serving. And it potentially makes it easy if in the future we want to support "Disabled".  WDYT?

mattmoor: It's either part of conformance or not.  The sentiments I was talking about in the doc weren't specific to Route at all, but all clients of the networking API.

pkg/apis/networking/v1alpha1/ingress_validation.go
```diff
@@ -166,3 +167,14 @@ func (t *IngressTLS) Validate(ctx context.Context) *apis.FieldError {
...
+// Validate inspects and validates HTTPOption object.
+func (t HTTPOption) Validate(ctx context.Context) (all *apis.FieldError) {
+	switch t {
+	case HTTPOptionEnabled, HTTPOptionDisabled, HTTPOptionRedirected:
+		break
+	default:
+		all = all.Also(apis.ErrInvalidValue(t, "httpOption"))
+	}
+	return all
+}
```
mattmoor, line 180: We definitely cannot add this, it means that ALL existing kingress are invalid.

pkg/apis/networking/v1alpha1/ingress_defaults.go
vagababov: Yeah see net-istio downstream failures too.

pkg/apis/networking/v1alpha1/ingress_validation.go
vagababov: Wouldn't defaulting solve this?

mattmoor: Defaulting only solves it if we change something in spec.  In the meantime, I believe status updates will fail to validate.  Regardless, we cannot default (yet) because it violates our ability to downgrade.

ZhiminXiang: Would it be OK to allow empty value of HTTPOption? I think if we allow empty string, then the Validate function should be fine, right?

mattmoor: Yeah, if it's there, we should validate it, but we shouldn't require it to be there.

PR #307, opened 2020-12-11
Add the config map setting for the gradual rollout

vagababov:
subj

/assign @tcnghia mattmoor

### DIFF
--- a/config/config-network.yaml
+++ b/config/config-network.yaml
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: "5e3df87d"
+    knative.dev/example-checksum: "150914b9"
 data:
   _example: |
     ################################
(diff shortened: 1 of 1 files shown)


config/config-network.yaml
```diff
@@ -104,3 +104,7 @@ data:
     # 3. Redirected: The Knative ingress will send a 302 redirect for all
     # http connections, asking the clients to use HTTPS.
     httpProtocol: "Enabled"
+
+    # rolloutDuration contains the minimal duration in seconds over which the
+    # Configuration traffic targets are rolled out to the newest revision.
+    rolloutDuration: "0"
```
mattmoor, line 110: Does anything else have to be enabled in order to use this? 🤔

vagababov: Not right now.
If we think it's necessary, we can add another feature flag — but having flag to enable flag, whose default value is off, seems too much?

PR #313, opened 2020-12-16
Avoid recreating the logger in the ingress prober

markusthoemmes:
Since this lib is designed to be called in a reconciler, I think it is fair to assume that the context is already carrying the correct logger. Passing it around in the workItems also avoid recreating it over and over.

/assign @tcnghia @vagababov


tcnghia (approved):
/lgtm

tcnghia:
@markusthoemmes this looks good to me.

@vagababov seems there are new flakiness in downstream unit tests. I wonder if related to the rollout work?

vagababov:
No it seems to be `httpOption` :)

PR #326, opened 2021-01-06
Add Namespace field to ClusterDomainClaim

julz:
# Changes

- :gift: Adds `Spec.Namespace` to `ClusterDomainClaim`. 

This field has two purposes:

- It replaces the use of ownerReferences for ensuring that only one
DomainMapping can use a ClusterDomainClaim at a time. This avoids
using cross-namespace ownerReferences, which aren't supported.
- It allows Cluster Operators to delegate a particular domain name to a
namespace, allowing that namespace (and no other) to create DomainMappings
with a particular name (thereby preventing other namespaces squatting on or
intercepting the name).

/kind enhancement

First part of https://github.com/knative/serving/issues/10477 (and also to enable multi-tenant use of DomainMapping, see https://github.com/knative/serving/issues/9713).

**Release Note**

```release-note
Adds `spec.Namespace` to `ClusterDomainClaim` to control which namespace owns the ability to create a `DomainMapping` with the same name as the claim.
```

/assign @mattmoor @tcnghia


pkg/apis/networking/v1alpha1/domainclaim_types.go
```diff
@@ -52,3 +57,12 @@ type ClusterDomainClaimList struct {
...
 	Items []ClusterDomainClaim `json:"items"`
 }
+
+// ClusterDomainClaimSpec is the desired state of the ClusterDomainClaim.
+// Its only field is `namespace`, which controls which namespace currently owns
+// the ability to create a DomainMapping with the ClusterDomainClaim's name.
+type ClusterDomainClaimSpec struct {
+	// Namespace is the namespace which is allowed to create a DomainMapping
+	// using this ClusterDomainClaim's name.
+	Namespace string `json:"namespace"`
```
mattmoor, line 67: We should think about requiring this next release.

mattmoor (commented):
/lgtm
/approve

PR #329, opened 2021-01-11
As per discussion with @mattmoor keep drainer in pkg

vagababov:
So remove it from here

/assign @tcnghia mattmoor


mattmoor (commented):
/lgtm

PR #330, opened 2021-01-12
Add config to disable auto-creation of domain claims

julz:
# Changes

- :gift: Adds `autocreateClusterDomainClaims` flag to network configmap. 

  Used to disable automatic claiming of domain names in favour of explicit delegation of domain name ownership to namespaces via the spec.Namespace field of `ClusterDomainClaim` (this is important for multi-tenant environments which need to prevent 'squatting' or 'intercepting' domain names used in other namespaces but exposed via the shared ingress).

/kind enhancement

**Release Note**

```release-note
Adds `autocreateClusterDomainClaims` flag to network config map.
```

/assign @mattmoor @tcnghia

### DIFF
--- a/pkg/network.go
+++ b/pkg/network.go
@@ -116,6 +116,10 @@ const (
 	// constructing the Knative Route's tag names.
 	DefaultTagTemplate = "{{.Tag}}-{{.Name}}"
 
+	// AutocreateClusterDomainClaimsKey is the key for the
+	// AutocreateClusterDomainClaims property.
+	AutocreateClusterDomainClaimsKey = "autocreateClusterDomainClaims"
+
 	// AutoTLSKey is the name of the configuration entry
 	// that specifies enabling auto-TLS or not.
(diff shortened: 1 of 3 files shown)


pkg/network.go
```diff
@@ -116,6 +116,10 @@ const (
 	// constructing the Knative Route's tag names.
 	DefaultTagTemplate = "{{.Tag}}-{{.Name}}"
 
+	// AutocreateClusterDomainClaimsKey is the key for the
+	// AutocreateClusterDomainClaims property.
+	AutocreateClusterDomainClaimsKey = "autocreateClusterDomainClaims"
```
mattmoor, line 121: I don't like this name, but I don't have a better suggestion. 🤔

mattmoor (commented):
/lgtm
/approve

pkg/network.go
julz: same :(. If we can come up with something better before we make this "beta", that'd be nice.

PR #343, opened 2021-02-02
[WIP] Generalize status readiness from ingress to any object

devguyio:
Trying to see how to use status package outside ingress


vagababov:
Can it be probed the same way?

matzew:
@devguyio what's the status here? Also that it breaks some of downstream. Are we in a state to merge this and we can use it on the eventing-kafka repo ?

PR #348, opened 2021-02-05
Add ingress conformance test to ensure we do not add retries

julz:
# Changes

- :gift: Adds a conformance test to ensure we do not retry failed requests. This regressed in https://github.com/knative-sandbox/net-istio/pull/263 (fix in https://github.com/knative-sandbox/net-istio/pull/508), so worth adding a test.

/kind enhancement

/assign @nak3 @tcnghia

### DIFF
--- a/test/conformance/ingress/retry.go
+++ b/test/conformance/ingress/retry.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ingress
+
+import (
+	"context"
+	"net/http"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"knative.dev/networking/pkg/apis/networking/v1alpha1"
+	"knative.dev/networking/test"
+)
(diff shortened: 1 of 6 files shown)


test/conformance/ingress/retry.go
```diff
@@ -0,0 +1,77 @@
...
+	"net/http"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"knative.dev/networking/pkg/apis/networking/v1alpha1"
+	"knative.dev/networking/test"
+)
+
+// TestRetry verifies that the ingress does not retry failed requests.
+func TestRetry(t *testing.T) {
```
vagababov, line 30: it's more like `TestNoRetry`

vagababov (commented):
/lgtm

test/conformance/ingress/retry.go
julz: ha, true, guess I was thinking one day we may reintroduce the Retry field and this would just be the base case 🤷

test/conformance/ingress/retry.go
```diff
@@ -0,0 +1,77 @@
...
+	})
+
+	// First try - we expect this to fail, because we shouldn't retry
+	// automatically and the service only responds 200 on the _second_ access.
+	resp, err := client.Get("http://" + domain)
+	if err != nil {
+		t.Errorf("Error making GET request: %v", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusOK {
```
nak3, line 62:
```suggestion
	if resp.StatusCode != http.StatusServiceUnavailable {
```

Is it better to verify this with `http.StatusServiceUnavailable` as we know the test image returns 503?

nak3:
/lgtm

Just one nit comment but LGTM.

test/conformance/ingress/retry.go
julz: seems reasonable to me - fixed

nak3 (approved):
/lgtm

Thank you!

PR #349, opened 2021-02-08
Bump a few assorted dependencies to their latest versions

markusthoemmes:
As per title, just to keep us chasing the latest and greatests version. A bit of a monday cleanup!


vagababov (commented):
/lgtm

PR #352, opened 2021-02-09
Add a bit more coverage the stat collector

markusthoemmes:
As per title, just to nail the behavior of this component a bit more as I got anxious about a few edge cases.

/assign @vagababov @julz


vagababov (commented):
/lgtm

PR #354, opened 2021-02-11
Add header for signifying that requests should not be loadbalanced again

markusthoemmes:
Part of https://github.com/knative/serving/issues/10751.

This header is necessary to instruct Ingress (in this case Istio) to not route the request via their Host tables and loadbalance things but rather respect the included IP address and route the request directly.

/assign @nak3

### DIFF
--- a/pkg/network.go
+++ b/pkg/network.go
@@ -176,10 +176,10 @@ const (
 	// consistent.
 	VisibilityLabelKey = "networking.knative.dev/visibility"
 
-	// DirectLoadbalancingHeaderName is the name of a header that signifies to
-	// loadbalancers that the respective request should not be loadbalanced again but
-	// sent to the request's target directly.
-	DirectLoadbalancingHeaderName = "K-Direct-Lb"
+	// PassthroughLoadbalancingHeaderName is the name of the header that directs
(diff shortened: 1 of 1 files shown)


pkg/network.go
```diff
@@ -175,6 +175,11 @@ const (
 	// already using labels for domain, it probably best to keep this
 	// consistent.
 	VisibilityLabelKey = "networking.knative.dev/visibility"
+
+	// DirectLoadbalancingHeaderName is the name of a header that signifies to
+	// loadbalancers that the respective request should not be loadbalanced again but
+	// sent to the request's target directly.
+	DirectLoadbalancingHeaderName = "K-Direct-Lb"
```
vagababov, line 182:
```suggestion
	// DirectLoadbalancingHeaderName is the name of the header that directs
	// load balancers that the respective request should not be load balanced but
	// sent to the request's target directly.
	DirectLoadbalancingHeaderName = "K-Direct-Lb"
```

I've edited the text, how I'd have written it, but you can ignore some or all of it :-)

Finally, should we just call it `passthrough` rather than `direct`?

markusthoemmes: Done.

vagababov (approved):
/lgtm

PR #377, opened 2021-03-26
Add pod-addressability config to config-network

markusthoemmes:
# Changes

Ref https://github.com/knative/serving/issues/10751

For the pod-addressability work we need a setting available to both net-istio and knative-serving so that each can adjust their behavior accordingly. I first added this setting to net-istio, intending to add a second one to knative-serving, but that seems unnecessary if we can have just one in config-network, readable by both. So here we are.

If this merges, I'll roll back the istio specific config of course.

/assign @nak3 @ZhiminXiang @julz

### DIFF
--- a/config/config-network.yaml
+++ b/config/config-network.yaml
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: "c42b8304"
+    knative.dev/example-checksum: "15954d34"
 data:
   _example: |
     ################################
@@ -125,4 +125,7 @@ data:
     # enabled and thus direct-addressability is usually not possible.
     # Consumers like Knative Serving can use this setting to adjust their behavior
     # accordingly, i.e. to drop fallback solutions for non-pod-addressable systems.
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
     enable-mesh-pod-addressability: "false"
--- a/config/config-network.yaml
+++ b/config/config-network.yaml
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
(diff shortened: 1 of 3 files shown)


julz (approved):
seems reasonable to me.

config/config-network.yaml
```diff
@@ -119,3 +119,10 @@ data:
     # which need to control which namespace can use a particular domain name in
     # a domain mapping.
     autocreateClusterDomainClaims: "true"
+
+    # If true, networking plugins can add additional information to deployed
```
julz, line 123: we should call out that this is an alpha/experimental feature

pkg/network.go
```diff
@@ -298,6 +309,7 @@ func NewConfigFromMap(data map[string]string) (*Config, error) {
 		cm.AsString(TagTemplateKey, &nc.TagTemplate),
 		cm.AsInt(RolloutDurationKey, &nc.RolloutDurationSecs),
 		cm.AsBool(AutocreateClusterDomainClaimsKey, &nc.AutocreateClusterDomainClaims),
+		cm.AsBool(EnableMeshPodAddressabilityKey, &nc.EnableMeshPodAddressability),
 	); err != nil {
 		return nil, err
 	}
```
nak3, line 315: As `EnableMeshPodAddressability` is only work with net-istio, should we add a validation for the class like `EnableMeshPodAddressability && nc.DefaultIngressClass != IstioIngressClassName`
?

markusthoemmes: Hmmmm, I **do** see your point but I'm wondering if it'll hamper testing of the feature. You could for example imagine of people using istio for some services and not as the default and we'd still want to set this I believe.

On the other hand, I don't see people misusing this just yet :thinking:. I think I'd lean towards keeping it open for now.

nak3: I understand and that makes sense. Let's keep as it is.

nak3 (approved):
/lgtm

PR #385, opened 2021-04-03
Append headers to TestRule to verify we hit expected backend

nak3:
`TestRule` creates two rules to confirm each rule work fine.
But we are not sure if test requests hit the correct backend - `fooName`
svc or `barName` svc by current code.

Hence this patch appends header to verify we hit expected backend

/kind enhancement

**Release Note**

```release-note
NONE
```


tcnghia (approved):
/lgtm

PR #396, opened 2021-04-20
Allow the prober to cancel probing by key

markusthoemmes:
# Changes

This adds the possibility to not only stop probing an Ingress by passing the ingress object, but also by passing just the namespace + name pair identifying the Ingress. See https://github.com/knative-sandbox/net-kourier/pull/497 for an example as to why this is necessary: When the object is already deleted, we only know the key.

/assign @nak3 @julz


julz (approved):
/lgtm

nak3 (approved):
/lgtm

PR #399, opened 2021-04-26
Add a schema to all CRDs

markusthoemmes:
# Changes

Ref https://github.com/knative/serving/issues/912

Same update as for Serving essentially.


markusthoemmes:
/hold

For the pkg change.

markusthoemmes:
/unhold

This is done from my view, if we even want this, seeing as these are internal CRDs.

markusthoemmes:
/hold

Something's odd with the downstream tests...

PR #400, opened 2021-04-26
Add missing JSON tag to protocol type

markusthoemmes:
# Changes

As per title. Existing object in etcd will have `ProtocolType` (uppercase) as a key, but this should be backwards compatible as `Unmarshal` allows for case-insensitive matching

> To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match.

(https://golang.org/pkg/encoding/json/#Unmarshal)

/assign @vagababov


markusthoemmes:
/assign @julz 

For good measure.

vagababov (approved):
/lgtm

PR #414, opened 2021-05-24
Remove deprecated cert and key field

nak3:
This patch removes `DeprecatedServerCertificate` and
`DeprecatedPrivateKey`.

They are deprecatd more than a year ago https://github.com/knative/serving/pull/6340
and it should be alright to remove.

/cc @tcnghia @ZhiminXiang @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

Issue #298, opened 2020-12-10
[HTTP Option support] Add conformance test for HTTP option


ZhiminXiang:
We need conformance test for HTTP option behavior to ensure the downstream implementations do the right thing.

Full design: https://docs.google.com/document/d/1yVOGo-j1hI7oCpuWVIi5iBJwX1wKbOzWaV0QNJxCbaY

ZhiminXiang:
/assign @nak3 
looks like the PR is ready but it is currently blocked by net-istio work.

PR #415, opened 2021-05-24
Add TestHTTPOption to verify HTTPOption field

nak3:
This patch adds `TestHTTPOption` which verifies that the Ingress
properly handles `HTTPOption` field.

Fix https://github.com/knative/networking/issues/298


nak3:
/hold

This is ready for review but I guess it should be hold until ingresses like istio, contour, kourier implements it :thinking:

test/conformance/ingress/httpoption.go
```diff
@@ -0,0 +1,117 @@
...
+		httpOption v1alpha1.HTTPOption
+		codes      []int
+	}{{
+		name:       "enabled",
+		httpOption: v1alpha1.HTTPOptionEnabled,
+		codes:      []int{http.StatusOK},
+	}, {
+		name:       "redirected",
+		httpOption: v1alpha1.HTTPOptionRedirected,
+		codes:      []int{http.StatusMovedPermanently /*301*/, http.StatusFound /*302*/},
```
julz, line 48: Really just out of interest: do we ever/would we ever/should we ever return a 302 here? My (very shaky) understanding was a http->https redirect would/should pretty much always be 301

nak3: Good catch. Yes, agree. 301 is almost always used for http to https redirect. 
I updated this PR as well as comment in config-network by https://github.com/knative/networking/pull/416.

test/conformance/ingress/httpoption.go
```diff
@@ -0,0 +1,110 @@
...
+	}, {
+		name:       "redirected",
+		httpOption: v1alpha1.HTTPOptionRedirected,
+		code:       http.StatusMovedPermanently,
+	}}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// net-istio cannot support parallel option as one HTTPOption effects globally.
+			// https://github.com/knative-sandbox/net-istio/issues/637
```
julz, line 54: Checking my understanding here: istio supports this, but it flips the setting for all ingresses back and forth each time it receives one? If so does that mean running this test will affect the whole cluster it runs against temporarily (because I think some people might run the tests against production clusters to test after deploying, so that may not be good..)?

nak3: > istio supports this, but it flips the setting for all ingresses back and forth each time it receives one?

Yes, exactly. Istio's HTTP is global. I created the issue https://github.com/knative-sandbox/net-istio/issues/637

> If so does that mean running this test will affect the whole cluster it runs against temporarily (because I think some people might run the tests against production clusters to test after deploying, so that may not be good..)?

Yes, that's also true. Perhaps we should NOT comment out this and let net-istio fix the issue  https://github.com/knative-sandbox/net-istio/issues/637.

julz: yeh, I think we may need to fix net-istio before landing this, unfortunately

nak3: And this conformance test should add an extra test that having multiple Ingress with different HTTPOption at the same time.

ZhiminXiang:
The fix in net-istio has been checked in. Is this PR ready for review?

nak3:
/hold cancel

Yes, it is ready for review. Although Contour and Kourier will need to skip this test as they don't implement it yet, I think it is fine to merge for Istio.

One of commits https://github.com/knative/networking/pull/415/commits/db1f702fcfefd008e7ccd5ca63430f48b1f9bd82 is reverted as https://github.com/knative/networking/pull/453 updated.

PR #433, opened 2021-06-04
Update Knative dependency for release-0.23

chizhg:


nak3:
I'm not sure about why this needs to be backport but https://github.com/knative/serving/pull/11465 merged the same fix so it is alright to merge here? :sweat_smile:

PR #437, opened 2021-06-10
Remove overrideInternalScheme and change semantics to defaulting

markusthoemmes:
# Changes

As per title. 

1. It was pointed out by @julz that semantics are not actually as described. We're going to leave AutoTLS alone for now. If people are using it, there's a good chance they want the semantics it provides and not muck with it, I think.
2. I wasn't sure about the usefulness of `overrideInternalScheme`, so I dropped it for now until somebody brings a use-case that actually needs it.

### DIFF
--- a/config/config-network.yaml
+++ b/config/config-network.yaml
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: "8f56afee"
+    knative.dev/example-checksum: "3e010c60"
 data:
   _example: |
     ################################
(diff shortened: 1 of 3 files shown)


config/config-network.yaml
```diff
@@ -134,7 +134,5 @@ data:
     # This can be used for making Knative report all URLs as "HTTPS" for example, if you're
     # fronting Knative with an external loadbalancer that deals with TLS termination and
     # Knative doesn't know about that otherwise.
+    # AutoTLS scheme semantics are left untouched.
```
julz, line 137: I'm going to be super annoying now, but just to make sure we explicitly think about this before it's hard to change: should this field be something like "DefaultExternalScheme" since it doesn't really override in this case?

(This way the docs wouldn't need to specify that it gets overridden by autotls but could just say it's the default value when autotls is not set)

config/config-network.yaml
```diff
@@ -134,7 +134,5 @@ data:
...
     overrideExternalScheme: ""
```
julz, line 137: Since this only really changes the default in the noautotls case now, does it make sense to have this default to "http" and thereby avoid the if OverrideExternalScheme == "" bits in the code?

markusthoemmes: If we go with the `default` you mentioned above, then yes! Imma think about it on my walk to lunch.

PR #453, opened 2021-06-24
Set default HTTPOption to Enabled for conformance test

ZhiminXiang:


test/conformance/ingress/util.go
```diff
@@ -686,6 +686,8 @@ func CreateIngress(ctx context.Context, t *testing.T, clients *test.Clients, spe
 		t.Fatalf("Invalid ingress %q: %v", ingName, err)
 	}
 
+	setDefaultsForTest(ing)
```
nak3, line 689: Do we have a plan to add a default value to `ing.Spec.HTTPOption` by `SetDefaults()`?
 If so, we should add `TODO:` comment.

ZhiminXiang: Yes. I think we should set the default value in SetDefaults in the future. I added a TODO here.

PR #457, opened 2021-07-05
Remove deprecated label `networking.internal.knative.dev/disableWildcardCert`

nak3:
/kind cleanup

As per title, the label was removed in serving repo https://github.com/knative/serving/commit/3d23011278339b467b280b8a3ed361e7ab7b5b8f,
so it should be safe to remove.


markusthoemmes (approved):
/lgtm
/approve

PR #465, opened 2021-07-16
[wip] split up knative.dev/networking/pkg

dprotaso:
- split up pkg into smaller packages
- drop corev1 use from pkg/config


dprotaso:
rebased in another PR

PR #476, opened 2021-08-04
Use actual cluster domain for TestRewriteHost

nak3:
The cluster domain is not always `cluster.local`. It could be a value defined by users.

So this patch replace the hard-coded `cluster.local` with `ClusterSuffix` via command line option.

P.S. I wondered why current test works and realized that the local domain is not used as https://github.com/knative/networking/blob/main/test/conformance/ingress/util.go#L553 actually.
So we can also use `privateHostName := privateServiceName + "." + test.ServingNamespace + ".svc"`.

### DIFF
--- a/test/conformance/ingress/rewrite.go
+++ b/test/conformance/ingress/rewrite.go
@@ -24,7 +24,6 @@ import (
 	"knative.dev/networking/pkg/apis/networking"
 	"knative.dev/networking/pkg/apis/networking/v1alpha1"
 	"knative.dev/networking/test"
-	"knative.dev/pkg/network"
 )
 
 // TestRewriteHost verifies that a RewriteHost rule can be used to implement vanity URLs.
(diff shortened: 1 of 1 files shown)


test/conformance/ingress/rewrite.go
```diff
@@ -34,7 +34,7 @@ func TestRewriteHost(t *testing.T) {
 	name, port, _ := CreateRuntimeService(ctx, t, clients, networking.ServicePortNameHTTP1)
 
 	privateServiceName := test.ObjectNameForTest(t)
-	privateHostName := privateServiceName + "." + test.ServingNamespace + ".svc.cluster.local"
+	privateHostName := privateServiceName + "." + test.ServingNamespace + ".svc." + test.NetworkingFlags.ClusterSuffix
```
tcnghia, line 37: Can we use https://github.com/knative/pkg/blob/3a2ae6db70971afbcaf1e1c3826cbc9a5d6c718e/network/domain.go#L40 instead?

nak3: It is not possible unfortunately.
`GetServiceHostname()` reads `/etc/hosts` so it works within a pod like activator code, queue-proxy code, etc. But here is a test client so it seems to read the `/etc/hosts` in the host machine which kicked test.  

I wondered that we can drop the `--cluster-suffix` option, but I understand why we cannot drop it :sweat_smile:

tcnghia (approved):

PR #482, opened 2021-08-19
Revert "Add a new test case to request unexpected hash (#468)"

nak3:
This reverts commit 078ce258cc65a91760bc7f5bcbe6ca471c81e4ad.

The new test case was wrong.
The ingress overrides the correct value only when `override` is set in
`K-Network-Hash` header, which is already covered by  "kingress generates hash" test.

/cc @tcnghia @ZhiminXiang @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

PR #494, opened 2021-09-02
Allow 502 codes as well as 503 to indicate mesh errors

julz:
We've had at least 2 reports now that in some cases with mesh enabled we see 502 codes instead of 503s, so seems to make sense to broaden the heuristic to accept those when considering whether we should fall back to mesh mode. In the past we accepted _any_ error to cause mesh fallback, so broadening seems pretty reasonable.

See https://github.com/knative/serving/issues/11877.

/assign @markusthoemmes @nak3 
~~/hold to see https://github.com/knative/serving/pull/11900 pass~~


julz:
tests in https://github.com/knative/serving/pull/11900 look good.

/unhold

PR #508, opened 2021-09-10
Create certificate pool for each test client

nak3:
Current cert pool is defined as a global variable `rootCAs`:

https://github.com/knative/networking/blob/release-0.25/test/conformance/ingress/util.go#L60

This was no problem before as ingress conformance test had only one TLS test.
But recently TestHTTPOption also starts using TLS client, so the global cert pool
causes a race issue like:

https://github.com/knative-sandbox/net-istio/pull/761#issuecomment-916586023
https://github.com/knative-sandbox/net-kourier/pull/636#issuecomment-916710041

To fix it, this patch changes to create a cert pool for each client and remove
the global one.

/kind cleanup

/cc @ZhiminXiang @tcnghia @markusthoemmes

### DIFF
--- a/test/conformance/ingress/httpoption.go
+++ b/test/conformance/ingress/httpoption.go
@@ -27,6 +27,14 @@ import (
 	"knative.dev/networking/test"
 )
 
+// testClient is a http client and expected status code to verify Ingress.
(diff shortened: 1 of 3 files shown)


markusthoemmes (approved):
/lgtm
/approve

Nice!

PR #519, opened 2021-09-28
Set FlushInterval to zero

julz:
As of go1.16 FlushInterval [works properly with streaming/websocket requests](https://go-review.googlesource.com/c/go/+/260637/), which means we no longer need the workaround of setting it to a low number of milliseconds.

See https://github.com/knative/serving/pull/11688 for the tests passing in serving with default FlushInterval.

/assign @markusthoemmes

### DIFF
--- a/pkg/network.go
+++ b/pkg/network.go
@@ -25,7 +25,6 @@ import (
 	"net/url"
 	"strings"
 	"text/template"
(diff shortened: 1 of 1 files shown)


markusthoemmes (approved):
/lgtm
/approve

PR #527, opened 2021-10-07
Allow to skip non-alpha/beta tests

nak3:
The `--skip-tests` flag was introduced to skip alpha or beta tests
before and it does not allow to skip stable test.

But now, to support gateway-api, we need to skip some stable tests as some
features are not supported even by the upstream.


nak3:
/cc @dprotaso @ZhiminXiang @carlisia 

This change needs net-gateway-api controller as it cannot pass some tests like https://github.com/knative-sandbox/net-gateway-api/issues/20 which is not supported by Gateway API CRD yet.

PR #562, opened 2021-11-16
Introduce kebab-case annotation variants 

dprotaso:
Introduce alternative annotation with kebab-case to be consistent with other parts of Knative Serving.

Serving will accept these variants on public resources (Service, Route) and use the older variants for now since those are more widely supported.

### DIFF
--- a/pkg/apis/networking/metadata_validation.go
+++ b/pkg/apis/networking/metadata_validation.go
@@ -26,9 +26,15 @@ import (
 
 var (
 	allowedAnnotations = sets.NewString(
-		DisableAutoTLSAnnotationKey,
+		IngressClassAnnotationKey,
 		CertificateClassAnnotationKey,
+		DisableAutoTLSAnnotationKey,
 		HTTPOptionAnnotationKey,
+
+		IngressClassAnnotationAltKey,
+		CertificateClassAnnotationAltKey,
+		DisableAutoTLSAnnotationAltKey,
+		HTTPProtocolAnnotationKey,
 	)
 )
 
--- a/pkg/apis/networking/register.go
+++ b/pkg/apis/networking/register.go
@@ -20,29 +20,6 @@ const (
 	// GroupName is the name for the networking API group.
 	GroupName = "networking.internal.knative.dev"
 
-	// IngressClassAnnotationKey is the annotation for the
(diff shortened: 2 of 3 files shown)


pkg/apis/networking/metadata_validation.go
```diff
@@ -29,6 +29,10 @@ var (
 		DisableAutoTLSAnnotationKey,
 		CertificateClassAnnotationKey,
 		HTTPOptionAnnotationKey,
+
+		DeprecatedCertificateClassAnnotationKey,
+		DeprecatedDisableAutoTLSAnnotationKey,
+		DeprecatedHTTPOptionAnnotationKey,
```
dprotaso, line 35: TODO: why isn't ingress class annotation here?

pkg/apis/networking/register.go
```diff
@@ -33,37 +56,30 @@ const (
 	// annotation value for the Ingress it uses.  Based on such
 	// value a different reconciliation logic may be used (for examples,
 	// Istio-based Ingress will reconcile into a VirtualService).
-	IngressClassAnnotationKey = "networking.knative.dev/ingress.class"
+	IngressClassAnnotationKey = PublicGroupName + "/ingress-class"
```
nak3, line 59: This annotation (`ingress.class`) was aligned with Kubernetes's ingres class annotation `kubernetes.io/ingress.class`.
But kubernetes started using `spec.ingressClassName` so we might be better to consider to use `spec` in the future.
https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

dprotaso:
/hold cancel

@nak3 I've simplified my approach here. It's easier if the serving controller just transforms these kebab-case annotations to the older format. This gives us time for downstream implementations to adopt the newer annotation case. 

This satisfies https://github.com/knative/serving/issues/12103 but still gives us lots of time to get the necessary changes to the downstream cert/ingress providers.

nak3:
Sounds good. So, net-* downstream repos do not need to adopt this change for a while.  Then, Is it possible to see the POC code on serving repo which uses this change?

dprotaso:
Serving should accept these annotations once this merges since it’s using
the helpers already

https://github.com/knative/serving/pull/12307

PR #564, opened 2021-11-16
drop used stubs

dprotaso:


dprotaso:
rebased so we get the go1.17 workflow

PR #567, opened 2021-11-17
include helpers to lookup annotations

dprotaso:
this should enable supporting multiple keys

### DIFF
--- a/pkg/apis/networking/register.go
+++ b/pkg/apis/networking/register.go
@@ -88,3 +88,19 @@ var (
 	// DefaultRetryCount will be set if Attempts not specified.
 	DefaultRetryCount = 3
 )
+
+func GetIngressClass(annotations map[string]string) (val string) {
+	return annotations[IngressClassAnnotationKey]
+}
+
+func GetCertificateClass(annotations map[string]string) (val string) {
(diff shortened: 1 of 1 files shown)


nak3:
I'm a little bit confused. It will become https://github.com/knative/networking/pull/562/files#diff-da8358898a61a0564b9d8d16c9a08bfe7acbfea66006fb6753b0f44b8c20ba1cR117-R163 eventually?

dprotaso:
Switching to the helper will make it easier to transition without breaking things.

I don't think we want to drop support for the old casing unless we perform some migration.

nak3:
Thank you. I understand what I misunderstood. The function in [pull/562](https://github.com/knative/networking/pull/562) uses `pkg/configmap.Parse` and `configmap.AsString` but it just parse annotations, not configmap` at all.
I misunderstood as how annotation parse will be changed to configmap parse :sweat:

dprotaso:
That other PR is a WIP - i realize the code is simply enough so I dropped
the use of the configmap package

dprotaso:
*in this PR

dprotaso:
But yeah I was using that package to parse values from the annotations

PR #588, opened 2021-12-06
Simplify the HTTP Proxy test image a bit

markusthoemmes:
# Changes

Basically the same as https://github.com/knative/serving/pull/12399. We didn't have the DNS cache stuff in Serving, so maybe this is good enough?


nak3:
/lgtm

Yeah, I wonder why DNS cache stuff was added :thinking:

PR #605, opened 2022-01-20
Add transport option

nak3:
We used to have the transport option before:

https://github.com/knative/networking/blob/release-1.0/test/test_images/httpproxy/httpproxy.go#L107-L110

but it was dropped and it looks like it caused visibility test flake.

This patch tries to verify the fix.

/cc @dprotaso @carlisia @ZhiminXiang


nak3:
Tested the change by net-istio and net-contour.
https://github.com/knative-sandbox/net-istio/pull/852
https://github.com/knative-sandbox/net-contour/pull/709

PR #608, opened 2022-01-26
Add `activator-ca` and `activator-name` keys in `config-network`

nak3:
This pach adds `activator-ca` and `activator-name` keys in `config-network`.

Part of https://github.com/knative-sandbox/net-kourier/issues/750
https://github.com/knative-sandbox/net-kourier/pull/761 demonstrates how it works.

/cc @evankanderson @julz @carlisia

PR #637, opened 2022-03-18
Add validation for activator-ca and activator-san

nak3:
This is a follow up for https://github.com/knative/networking/pull/608.

This patch adds a validation to disallow `activator-ca` without `activator-san`.
I think we can relax the validation in the future if users want to skip SAN validation
but the initial implementation should be strict.

/cc @evankanderson @ZhiminXiang @carlisia


evankanderson (commented):
/lgtm
/approve

PR #648, opened 2022-04-06
Add certificates config keys in config-network

nak3:
This patch adds the following certificate variables:

- `activator-server-certs`
- `queue-proxy-ca`
- `queue-proxy-san`
- `queue-proxy-server-certs`

It is similar to https://github.com/knative/networking/pull/608.

https://github.com/knative/serving/pull/12815 and https://github.com/knative/serving/pull/12770
verified the change.

/cc @evankanderson

### DIFF
--- a/config/config-network.yaml
+++ b/config/config-network.yaml
@@ -22,7 +22,7 @@ metadata:
     app.kubernetes.io/component: networking
     app.kubernetes.io/version: devel
   annotations:
-    knative.dev/example-checksum: "7c86cb6a"
+    knative.dev/example-checksum: "65b0cddb"
 data:
   _example: |
     ################################
@@ -188,3 +188,35 @@ data:
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
     activator-san: ""
+
+    # The server certificates to serve the TLS traffic from ingress to activator.
+    # It is specified by the secret name, which has the "tls.crt" and "tls.key" data field.
+    # Use an empty value to disable the feature (default).
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    activator-server-certs: ""
+
+    # The CA public certificate used to sign the queue-proxy TLS certificate.
+    # It is specified by the secret name, which has the "ca.crt" data field.
+    # Use an empty value to disable the feature (default).
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    queue-proxy-ca: ""
+
+    # The SAN (Subject Alt Name) used to validate the activator TLS certificate.
+    # It must be set when "queue-proxy-ca" is specified.
+    # Use an empty value to disable the feature (default).
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    queue-proxy-san: ""
+
+    # The server certificates to serve the TLS traffic from activator to queue-proxy.
+    # It is specified by the secret name, which has the "tls.crt" and "tls.key" data field.
+    # Use an empty value to disable the feature (default).
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    queue-proxy-server-certs: ""
--- a/pkg/network.go
+++ b/pkg/network.go
@@ -196,6 +196,21 @@ const (
 
 	// ActivatorSANKey is the config for the SAN used to validate the activator TLS certificate.
 	ActivatorSANKey = "activator-san"
+
+	// ActivatorServerCertKey is the config for the secret name, which stores certificates
+	// to serve the TLS traffic from ingress to activator.
+	ActivatorServerCertKey = "activator-server-certs"
+
+	// QueueProxyCAKey is the config for the secret name, which stores CA public certificate used
+	// to sign the queue-proxy TLS certificate.
+	QueueProxyCAKey = "queue-proxy-ca"
+
+	// QueueProxySANKey is the config for the SAN used to validate the queue-proxy TLS certificate.
+	QueueProxySANKey = "queue-proxy-san"
+
+	// QueueProxyServerCertKey is the config for the secret name, which stores certificates
+	// to serve the TLS traffic from activator to queue-proxy.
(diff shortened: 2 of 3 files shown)


config/config-network.yaml
```diff
@@ -188,3 +188,35 @@ data:
...
+    #       for now. Use with caution.
+    queue-proxy-san: ""
+
+    # The server certificates to serve the TLS traffic from activator to queue-proxy.
+    # It is specified by the secret name, which has the "tls.crt" and "tls.key" data field.
+    # Use an empty value to disable the feature (default).
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    queue-proxy-server-certs: ""
```
evankanderson, line 222: Shouldn't queue-proxies in different namespaces get different SANs and Certs?

config/config-network.yaml
```diff
@@ -188,3 +188,35 @@ data:
...
     #       for now. Use with caution.
     activator-san: ""
+
+    # The server certificates to serve the TLS traffic from ingress to activator.
+    # It is specified by the secret name, which has the "tls.crt" and "tls.key" data field.
+    # Use an empty value to disable the feature (default).
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
+    #       for now. Use with caution.
+    activator-server-certs: ""
```
evankanderson, line 198: How about `activator-cert-secret` if we need this to be configurable? (I'm okay with it being configurable, though I'm not sure how useful that feature might be vs simply having a hard-coded name.)

evankanderson (commented):
I think we want to have a pattern for queue-proxy SANs which incorporates at least the namespace name, and possibly also the revision name.

I'm less concerned about revision name, but it would be nice if we could use the SAN to prove that we don't "cross the streams" at the networking layer and dispatch a request to a queue-proxy at the wrong namespace.

config/config-network.yaml
nak3: I wanted make this configurable as activator can turn on the https serving or not based on the config as:

```golang
	if networkConfig.ActivatorServerCert != "" {
              // activator starts serving https.
        }
```

config/config-network.yaml
nak3: Yes, it should get the different SANs and certs. But for the alpha release, I assumed that:
- admin deploys secret (server certs) in each application namespace with the name specified in `queue-proxy-server-certs`.
- queue-proxy mounts the secret via volume mount.

The different SANs and certs are very difficult to implement and we need to consider about another traffic `ingress -> queue-proxy` in the future.

nak3:
For queue-proxy's SANs and certs, it needs to consider about the `ingress -> queue-proxy`, which we have a plan to implement in the future, right? I think it is difficult to implement it in this alpha release...

config/config-network.yaml
nak3: For the name of `activator-cert-secret` vs `activator-server-certs`, we already have the activator ca as `activator-ca` (without `-secert`) so it should be consistent?

config/config-network.yaml
evankanderson: It seems important that a user in one namespace not have access to the cert being used in a different namespace. SANs feel like the way to do that -- we can start with having the admins manually create the certs using something like Cert-Manager or by hand, and eventually enhance Knative to do this before GA.

config/config-network.yaml
evankanderson: I'd thought the `activator-ca` was the actually CA certificate (which doesn't need to be kept private, because it's a certificate, not keys).

I'd prefer to have the `-secret` name here, because it provides a clearer clue that this is a value that needs to be kept private as well.

evankanderson:
> For queue-proxy's SANs and certs, it needs to consider about the `ingress -> queue-proxy`, which we have a plan to implement in the future, right? I think it is difficult to determine/implement it with the namespace level in this alpha release...

I think the challenge for `ingress -> queue-proxy` is twofold:

1. Need some way to teach Ingresses about SANs and CAs associated with particular endpoints (i.e. a mix of `activator` and `queue-proxy` pods will have different SANs depending on the endpoint).
2. Need some way to teach individual Ingresses about per-SLS SANs, which means we probably need to put that information in KIngress, which means a "protocol feature bump" in a way that is harder than `activator -> queue-proxy`, where we control all the code (which is in Go).

I'd be happy with either hard-coding the SAN as equal to the internal cluster service name, or something like `{{namespace}}.knative`.

config/config-network.yaml
skonto: I think we need to validate SAN/certs at service creation time. It should be done for GA. For now its ok to test with shared certs I guess.

skonto:
> I think the challenge for ingress -> queue-proxy is twofold: 
...

I would also vote for moving the san/cert info in the ingress part.

> I'd be happy with either hard-coding the SAN as equal to the internal cluster service name, or something like {{namespace}}.knative.

SAN can be many things afaik. Could users use the service without the internal name (thinking out loud)?

evankanderson:
> > I think the challenge for ingress -> queue-proxy is twofold:
> > ...
> 
> I would also vote for moving the san/cert info in the ingress part.

I'm not sure I understand the comment here. Do you mean move the SAN and Cert info in the KIngress, or into a different phase of the implementation.

> > I'd be happy with either hard-coding the SAN as equal to the internal cluster service name, or something like {{namespace}}.knative.
> 
> SAN can be many things afaik. Could users use the service without the internal name (thinking out loud)?

SAN is a field in the CA certificate; I don't think we need to align those values with specific resolvable DNS addresses. I think I'm suggesting using the dnsNames field here only because that seems to be supported by cert-manager, but I'm imagining that we'd have a custom validator and wouldn't want other systems using the internal name and attempting validation.

nak3:
@skonto @evankanderson Is it possible to merge this into 1.4 :sweat_smile:

evankanderson:
> @skonto @evankanderson Is it possible to merge this into 1.4 😅

I'd like to understand the plan for queue-proxy certs. Is it to copy a shared cert to each namespace for 1.4, and then to replace them (and the related config options) in 1.5?

nak3:
> Is it to copy a shared cert to each namespace for 1.4, and then to replace them (and the related config options) in 1.5?

Yes, I assumed that.

evankanderson:
I think this probably means that the feature will be pre-alpha for 1.4, but as long as we're fine with throwing out some or all of users configuration when we ship 1.5.

/lgtm

PR #668, opened 2022-05-11
Add label selector for secrets

skonto:
# Changes

- Needed for https://github.com/knative-sandbox/net-istio/pull/920, as suggested here: https://github.com/knative/pkg/pull/2516#pullrequestreview-968466191
- Need to merge this first.

/kind enhancement

### DIFF
--- a/pkg/apis/networking/register.go
+++ b/pkg/apis/networking/register.go
@@ -23,7 +23,7 @@ const (
 	GroupName = "networking.internal.knative.dev"
 
 	// CertifcateUIDLabelKey is used to specify a label selector for informers listing ingress secrets.
-	CertifcateUIDLabelKey = GroupName + "/certificate-uid"
+	CertificateUIDLabelKey = GroupName + "/certificate-uid"
 
 	// IngressLabelKey is the label key attached to underlying network programming
 	// resources to indicate which Ingress triggered their creation.
--- a/pkg/apis/networking/register.go
+++ b/pkg/apis/networking/register.go
@@ -23,7 +23,7 @@ const (
 	GroupName = "networking.internal.knative.dev"
 
 	// CertifcateUIDLabelKey is used to specify a label selector for informers listing ingress secrets.
-	CertifcateUIDLabelKey = GroupName + "/certficate-uid"
+	CertifcateUIDLabelKey = GroupName + "/certificate-uid"
 
 	// IngressLabelKey is the label key attached to underlying network programming
 	// resources to indicate which Ingress triggered their creation.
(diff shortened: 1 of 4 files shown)


test/conformance/ingress/util.go
```diff
@@ -985,7 +986,8 @@ func CreateTLSSecret(ctx context.Context, t *testing.T, clients *test.Clients, h
 			Name:      name,
 			Namespace: test.ServingNamespace,
 			Labels: map[string]string{
-				"test-secret": name,
+				"test-secret":              name,
+				pkg.SecretInformerLabelKey: "test",
```
skonto, line 990: This is dummy we only need this because of the fact that net-istio uses the tests in this repo.

pkg/network.go
```diff
@@ -211,6 +211,9 @@ const (
 	// QueueProxyCertKey is the config for the secret name, which stores certificates
 	// to serve the TLS traffic from activator to queue-proxy.
 	QueueProxyCertKey = "queue-proxy-cert-secret"
+
+	// SecretInformerLabelKey is used to specify a label selector for informers listing ingress secrets.
+	SecretInformerLabelKey = "certificate.networking.knative.dev"
```
dprotaso, line 216: `SecretInformerLabelKey` is too generic. 

If from the net-certmanager PR if we end up using the certificate UID as the label then

```go
CertifcateUIDLabelKey = "networking.knative.dev/certficateUID"
````

**Review by AngeloDanducci — COMMENTED** [review]
Agree with Dave - might change the order of merges but UID seems like a safe bet.

File: pkg/network.go (hunk 2 above)
Comment (dprotaso) [reply pkg/network.go:216]: the label could live here: https://github.com/knative/networking/blob/main/pkg/apis/networking/register.go

**Comment by skonto** [comment]
The ordering should not change but will rename the key.

File: pkg/network.go (hunk 2 above)
Comment (skonto) [reply pkg/network.go:216]: Could be yes will change to that np.

File: pkg/apis/networking/register.go [hunk 3]
```diff
@@ -22,6 +22,9 @@ const (
 	// GroupName is the name for the networking API group.
 	GroupName = "networking.internal.knative.dev"
 
+	// CertifcateUIDLabelKey is used to specify a label selector for informers listing ingress secrets.
+	CertifcateUIDLabelKey = GroupName + "/certficate-uid"
```
dprotaso, line 26:
```suggestion
	CertificateUIDLabelKey = GroupName + "/certificate-uid"
```

skonto:
@dprotaso my bad fixed the typo

dprotaso:
there were two typos - in the literal and the identifier 

`CertifcateUIDLabelKey` => `CertificateUIDLabelKey`

skonto:
My bad thanks for the patience.

skonto:
@dprotaso ready thanks for the review.

PR #670, opened 2022-05-13
Split up the `networking/pkg` to separate packages

dprotaso:
Replaces: https://github.com/knative/networking/pull/465

Part of: https://github.com/knative/serving/issues/9957

- bump go.mod to 1.17 and re-vendor
- pull out config-network parsing to it's own package
- split up knative.dev/pkg/network into seperate sub-packages
- use new packages

Note: I created `type` and `func` aliases so this change should be backwards compatible


pkg/deprecated_http.go
```diff
@@ -0,0 +1,116 @@
+/*
+Copyright 2018 The Knative Authors
```
ZhiminXiang, line 2: nit: 2022

ZhiminXiang (approved):
/lgtm

nak3:
I can some boiler plate error as https://github.com/knative/networking/runs/6425664316?check_suite_focus=true but LGTM in general.

As long as I checked [pkg/deprecated_config.go](https://github.com/knative/networking/pull/670/files#diff-46dca73442a5122c59293b8a5d9239060ee30bf87ca7850c127bc314692e35d9) can keep the compatibility. (If not, we should fix or revert :smiling_imp: )

PR #680, opened 2022-06-08
Support `internal-encryption` to deploy internal certificates automatically

nak3:
Currently users have to deploy certificates manually with several options such as
`activator-san`, `activator-ca`, `queue-proxy-ca` etc.

Such deployment and management of the certificates is a big burden for users.

Hence, this patch supports `internal-encryption` config to deploy
internal certificates automatically.

https://github.com/knative-sandbox/net-kourier/pull/855 and https://github.com/knative/serving/pull/13005 demonstrated this change.

### DIFF
--- a/config/config-network.yaml
+++ b/config/config-network.yaml
@@ -22,7 +22,7 @@ metadata:
     app.kubernetes.io/component: networking
     app.kubernetes.io/version: devel
   annotations:
-    knative.dev/example-checksum: "95478c0c"
+    knative.dev/example-checksum: "73d96d1b"
 data:
   _example: |
     ################################
@@ -173,7 +173,7 @@ data:
     # Knative doesn't know about that otherwise.
     default-external-scheme: "http"
 
-    # internal-encryption indicates whether internal trafic is encrypted or not.
+    # internal-encryption indicates whether internal traffic is encrypted or not.
     # If this is "true", the following traffic are encrypted:
     #  - ingress to activator
     #  - ingress to queue-proxy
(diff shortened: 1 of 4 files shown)


config/config-network.yaml
```diff
@@ -173,50 +173,12 @@ data:
...
-    # Use an empty value to disable the feature (default).
+    # internal-encryption indicates whether internal traffic is encrypted or not.
+    # If this is "true", the following traffic are encrypted:
+    #  - ingress to activator
+    #  - ingress to queue-proxy
+    #  - activator to queue-proxy
     #
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
-    activator-ca: ""
```
skonto, line 182: I guess all that will be in the `knative-serving-certs` and SAN is set to `data-plane.knative.dev`

evankanderson: Should the SAN include a per-namespace component?

evankanderson (commented):
I love the simplification in configuration! One concern -- I think we should have different SANs per-namespace, to prevent any sort of network-confusion attacks

evankanderson:
> > Note, I dropped the old configurations but the alpha feature can be dropped without notice https://knative.dev/docs/serving/configuration/feature-flags/#lifecyle
> 
> My initial thought was issue to mark the old configs as deprecated for a release and then drop them in the next one, but it looks like these flags were only added a couple of months ago in #648 , so don't think there's any problem with dropping.

I'm sort of in the same boat as Paul here -- I could go either way given how new this is and how hard the old way was to set up.

I think my larger / largest concern about keeping vs dropping values is whether it makes it hard to integrate from this repo to the `net-kourier` repo.

evankanderson:
(since the SAN is not actually specified in this PR, just mentioned by @skonto 's comment)

PR #712, opened 2022-09-26
WIP - proxy protocol probe mvp, needs discussion

AngeloDanducci:
# Changes

- Adds proxy protocol support when probing

/kind enhancement

Fixes
https://github.com/knative/serving/issues/13123
https://github.com/knative-sandbox/net-istio/issues/463

This is my first contribution to networking directly and I have a fair amount of questions, so I'll try and add tasks for questions that I think should be answered before this gets merged.

**New Dependency**

I've added a new [third party dependency](https://github.com/pires/go-proxyproto) that handles the proxy protocol header construction.

I've added the license file to the third party folder, the project is licensed under Apache 2.0 so I assume no issues here.

I've updated the go mod file, but I have not included a copy of the code in the vendor folder.

~~Should the vendor folder include a full copy or is the go mod update sufficient?~~

**Initial testing**

Initial testing scenario is as follows:

1. Knative serving installed normally via yaml
2. net-istio picking up this set of changes locally, use ko to apply this new config to my test cluster
3. Configure Magic DNS / sslip
4. [apply the proxy protocol envoy filter as seen in the istio documentation](https://istio.io/latest/docs/ops/configuration/traffic-management/network-topologies/#proxy-protocol)
5. create a new test service
6. observe that the test service becomes ready, and responds to normal network traffic

**Design Discussion**

The IP / port are required for the proxy protocol probe call.

The initial call is handled in [status when processing work items](https://github.com/knative/networking/blob/main/pkg/status/status.go#L347) and currently uses transport to create an opaque http roundtripper when performing the prober.do function.  

Process work item also had the raw pod ip / port available, so I expanded the [prober do function for ease of testing. ](https://github.com/knative/networking/compare/main...AngeloDanducci:networking:ad-proxyprotocol#diff-5248635f7bc24bf12f6f8d8320d937cbed8c2bd0f08f2a190729170324247209R106)

This works fine BUT the prober.Do function is [called in one more place](https://github.com/knative/networking/blob/main/pkg/prober/prober.go#L190) and this does not seem to have access to pod ip / port to add as arguments.

Given the above:

~~Is there a way to get pod IP / port [from within the do function](https://github.com/knative/networking/blob/main/pkg/prober/prober.go#L97)?  I tried a few things here to resolve the IP but the results appeared incorrect when using MagicDNS.  For example this is the service url from the test `test-ppp-1.default.30.28.114.84.sslip.io:80` which gets resolved as `30.28.114.84:80`, but the actual pod IP when given through argument was `172.28.208.132`.  I'm no network wizard so I may have missed something obvious here.~~
~~Alternatively would it make more sense to modify the offer or doasync functions so that the [do call from within the doasync function has the additional arguments?](https://github.com/knative/networking/blob/main/pkg/prober/prober.go#L190)~~

Refactored this to not support the offer/doAsync portion - they don't need to be supported as it's only used by serving's PA and that probes via private service.

Additional Discussion:

- [ ] When creating the proxy protocol header should we use the calling pod as the source IP?  
       It could have some default instead, representing any of these calls as being performed by a prober.  This call should only be a roundtrip, and the interesting sources are usually external.
- [ ] Where should the data for the proxy protocol flag live? I think sticking it in [here as a new item under data makes sense](https://github.com/knative/networking/blob/main/config/config-network.yaml).


AngeloDanducci:
cc @nak3

psschwei:
>     * [ ]  Should the vendor folder include a full copy or is the go mod update sufficient?

There's a `./hack/update-deps.sh` script that should take care of this for you

nak3:
> This works fine BUT the prober.Do function is called in one more place and this does not seem to have access to pod ip / port to add as arguments.

Do we need to support proxy protcol in the `doAsync()`'s `Do()` func? I think the `doAsync` is used by serving's PA only and it probes via private service not external so I wonder we don't need to support it.
If so, we can add the `DoWithProxyprotol()` func for the proxyprotol and use it in only `processWorkItem()`:

```go
if ctx.Value(ProxyProcolEnabled) {
        ok, err := prober.DoWithProxyProtol()
else {
        ok, err := prober.Do()
	...
}
```

nak3:
> Where should the data for the proxy protocol flag live? I think sticking it in [here as a new item under data makes sense](https://github.com/knative/networking/blob/main/config/config-network.yaml).

Yes, I agree. We should have it in config-network.

AngeloDanducci:
@nak3 could use a review when you get a chance

Let me know if there's a better way to consume the `config-network` values.
Needed to add a volume mount to support the current approach https://github.com/knative-sandbox/net-istio/pull/1028

Will need to test with Contour but I imagine it's the same as Istio - just need to consume the configmap.

AngeloDanducci:
Closing this with clean changes moved to here sans rebase: https://github.com/knative/networking/pull/738

PR #720, opened 2022-10-12
Fix lint error due to deprecated grpc.WithInsecure

nak3:
The new deps update started complaining about the deprecation as https://github.com/knative/networking/actions/runs/3231954423/jobs/5292030609

```
  Error: SA1019: grpc.WithInsecure is deprecated: use WithTransportCredentials and insecure.NewCredentials() instead. Will be supported throughout 1.x.  (staticcheck)
  Error: SA1019: grpc.WithInsecure is deprecated: use WithTransportCredentials and insecure.NewCredentials() instead. Will be supported throughout 1.x.  (staticcheck)
```
This patch fixes it.

/cc @lionelvillard


upodroid (approved):
LGTM

PR #721, opened 2022-10-20
fix http01 conformance tests

dprotaso:
LetsEncrypt no longer allows us to order certs for `example.com` it returns a HTTP 400 Bad Request.

This breaks the conformance tests as it expects an 'order' to be submitted (though it doesn't expect it to succeed)

### DIFF
--- a/test/conformance/certificate/http01/certificate.go
+++ b/test/conformance/certificate/http01/certificate.go
@@ -32,8 +32,8 @@ func TestHTTP01Challenge(t *testing.T) {
 	ctx, clients := context.Background(), test.Setup(t)
 
 	certDomains := [][]string{
-		{subDomain + ".knative.dev"},
-		{subDomain + "2.knative.dev", subDomain + "3.knative.dev"},
+		{subDomain + ".knative-test.dev"},
+		{subDomain + "2.knative-test.dev", subDomain + "3.knative-test.dev"},
 	}
 
 	for _, domains := range certDomains {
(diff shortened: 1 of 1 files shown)


dprotaso:
rebased so the downstream tests work (https://github.com/knative/networking/pull/722)

test/conformance/certificate/http01/certificate.go
```diff
@@ -32,8 +32,8 @@ func TestHTTP01Challenge(t *testing.T) {
 	ctx, clients := context.Background(), test.Setup(t)
 
 	certDomains := [][]string{
-		{subDomain + ".example.com"},
-		{subDomain + "2.example.com", subDomain + "3.example.com"},
+		{subDomain + ".knative.dev"},
```
dprotaso, line 35: @upodroid @kvmware @nak3 

Do you know if submitting HTTP01 requests to the subdomain of *.knative.dev exhausts some global quota that could affect knative.dev website from refreshing it's cert?

I'm assuming no because that could open up DoS attack - but asking just in case

upodroid: https://letsencrypt.org/docs/rate-limits/

I think these apply to both http01 and dns01 challenges

dprotaso: thanks for the input - I updated the domain to something we don't use `knative-test.dev`

nak3:
/lgtm

Thank you!

PR #722, opened 2022-10-20
fix downstream tests

dprotaso:
Downstream tests should be using the reusable action from knative/actions


nak3:
/lgtm

Thank you so much!

PR #747, opened 2023-01-03
add new field 'Domain' to kcert spec

KauzClay:
# Changes

- 🎁 
- Part of implementing a fix for https://github.com/knative-sandbox/net-certmanager/issues/214
- By allowing the KCert resource to have the domain, we can use it in net-certmanager to create a valid length CommonName in the Certificate
- Will be followed up by a PR to knative/serving to actually populate this field (see https://github.com/knative/serving/pull/13569)

/kind api-change

Part of Fix for https://github.com/knative-sandbox/net-certmanager/issues/214

**Release Note**

```release-note
N/A
```

**Docs**

```docs
N/A
```

### DIFF
--- a/pkg/apis/networking/v1alpha1/certificate_validation.go
+++ b/pkg/apis/networking/v1alpha1/certificate_validation.go
@@ -18,6 +18,7 @@ package v1alpha1
 
 import (
 	"context"
+	"strings"
 
 	"knative.dev/pkg/apis"
 )
@@ -44,5 +45,21 @@ func (spec *CertificateSpec) Validate(ctx context.Context) (all *apis.FieldError
 	if spec.SecretName == "" {
 		all = all.Also(apis.ErrMissingField("secretName"))
 	}
+
+	if spec.Domain != "" && len(spec.DNSNames) != 0 {
+		suffix := "." + spec.Domain
+		valid := false
+		for _, dnsName := range spec.DNSNames {
+			if strings.HasSuffix(dnsName, suffix) {
+				valid = true
+				break
+			}
+		}
+
+		if !valid {
+			all = all.Also(apis.ErrInvalidValue("domain", "domain must be a suffix of at least one entry in dnsNames"))
+		}
+	}
+
 	return all
 }
--- a/pkg/apis/networking/v1alpha1/certificate_validation_test.go
+++ b/pkg/apis/networking/v1alpha1/certificate_validation_test.go
@@ -33,6 +33,13 @@ func TestCertificateSpecValidation(t *testing.T) {
 		name: "valid",
 		cs: &CertificateSpec{
 			DNSNames:   []string{"host.example"},
+			SecretName: "secret",
+		},
+		want: nil,
+	}, {
+		name: "valid-with-domain",
+		cs: &CertificateSpec{
+			DNSNames:   []string{"host.example"},
 			Domain:     "example",
 			SecretName: "secret",
 		},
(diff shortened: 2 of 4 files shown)


KauzClay:
I started with making the Domain field required, but many tests fail. I think this is because net-certmanager tests don't know about it yet. I don't know how to make it required and make the tests pass.

Making it optional gets the tests to pass, but I don't know if that is the best end state.

KauzClay:
@evankanderson thanks for explaining, it makes sense why it should be optional then.

So, I added the `// +optional` and `omitempty` to the json tag. Is there anything else that needs to happen?

evankanderson (commented):
You may want to add a check that `spec.domain` is a suffix of at least one domain when it is set. See here if you want to enforce that:

https://github.com/knative/networking/blob/1cadfa6e5eba7b633c44cff27f813f8dd6fb0614/pkg/apis/networking/v1alpha1/certificate_validation.go#L31

pkg/apis/networking/v1alpha1/certificate_validation.go
```diff
@@ -44,5 +46,20 @@ func (spec *CertificateSpec) Validate(ctx context.Context) (all *apis.FieldError
 	if spec.SecretName == "" {
 		all = all.Also(apis.ErrMissingField("secretName"))
 	}
+
+	if spec.Domain != "" && len(spec.DNSNames) != 0 {
+		valid := false
+		for _, dnsName := range spec.DNSNames {
+			if strings.HasSuffix(dnsName, fmt.Sprintf(".%s", spec.Domain)) {
```
evankanderson, line 53: Rather than calling `fmt.Sprintf()` in the loop, do it once outside the loop.

Also, you should be able to use string `+`, rather than needing to pull in the `fmt` module:

```golang
if spec.Domain != "" {  // Already added an error for missing dnsNames
  domainSuffix := "." + spec.Domain
  err := apis.ErrInvalidValue(...)
  for _, dnsName := range spec.DNSNames {
    if strings.HasSuffix(dnsName, domainSuffix) {
      err = nil
      break
    }
  }
  all = all.Also(err)  // nil FieldError is a no-op
}
```

pkg/apis/networking/v1alpha1/certificate_validation_test.go
```diff
@@ -33,30 +33,42 @@ func TestCertificateSpecValidation(t *testing.T) {
 		name: "valid",
 		cs: &CertificateSpec{
 			DNSNames:   []string{"host.example"},
+			Domain:     "example",
```
evankanderson, line 36: Do you need a case without `Domain` set? (I think you do)

pkg/apis/networking/v1alpha1/certificate_validation.go
KauzClay: ah right, I'll fix that

pkg/apis/networking/v1alpha1/certificate_validation_test.go
```diff
@@ -36,27 +36,46 @@ func TestCertificateSpecValidation(t *testing.T) {
 			SecretName: "secret",
 		},
 		want: nil,
+	}, {
+		name: "valid-with-domain",
+		cs: &CertificateSpec{
+			DNSNames:   []string{"host.example"},
```
evankanderson, line 42: Not strictly necessary; you could add `"host.foo"` to DNSNames to test both paths through the loop, but I'm pretty confident with the coverage overall.

evankanderson (commented):
/lgtm
/approve

Thanks for the iterations!

PR #778, opened 2023-03-03
Add internal trust flag to config

### LINKED ISSUE #777: Add a flag to control future trust support
Current tls implementation support internal-encryption without trust
We need to have a roadmap for one with mutual trust which seems achievable using a number of future steps. 

Add flag for InternalTrust bool in knative.dev/networking/pkg/config/config.go so we can start alpha it.

/assign davidhadas

davidhadas:
Fixes #777


config/config-network.yaml
```diff
@@ -182,3 +183,14 @@ data:
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
     internal-encryption: "false"
+
+    # internal-trust indicates whether internal traffic is encrypted and mutual trust is achieved.
+    # Using internal-trust protect against MITM attacks
```
evankanderson, line 188: I think getting SANs fixed to be distinct per namespace also prevents MITM attacks, no?  (Now, it doesn't authenticate _clients_, so you can potentially bypass the KIngress implementation, but that's not a MITM attack...)

config/config-network.yaml
```diff
@@ -182,3 +183,14 @@ data:
...
+    # If this is "true", the following traffic are encrypted:
```
evankanderson, line 189: I think this is "the following traffic verifies client authentication with mutual TLS", not simply encrypted.

pkg/config/config.go
```diff
@@ -251,8 +255,11 @@ type Config struct {
 	// not enabled. Defaults to "http".
 	DefaultExternalScheme string
 
-	// DefaultExternal specifies whether internal traffic is encrypted or not.
+	// InternalEncryption specifies whether internal traffic is encrypted or not.
 	InternalEncryption bool
+
+	// InternalTrust specifies whether internal traffic is encrypted with mutual trust between peers.
```
evankanderson, line 261: Does this mean that `internal-trust: true` and `internal-encryption: false` means that encryption is enabled?

evankanderson (commented):
I'm going to continue to object to the idea that verifying a server SAN and CA does not provide MITM protection, but I'm happy to enable the `internal-trust` boolean for future mutual TLS.

config/config-network.yaml
davidhadas: I agree

davidhadas: For this SANs need to be meaningful. I.e. as a minimum a different SAN per namespace

pkg/config/config.go
davidhadas: This is what I mean.

config/config-network.yaml
```diff
@@ -22,163 +22,6 @@ metadata:
...
...and\n# Knative doesn't know about that otherwise.\ndefault-external-scheme: \"http\"\n\n# internal-encryption indicates whether internal traffic is encrypted or not.\n# Using internal-encryption does not protect against MITM attacks\n# If this is \"true\", the following traffic are encrypted:\n#  - ingress to activator\n#  - ingress to queue-proxy\n#  - activator to queue-proxy\n#\n# NOTE: This flag is in an alpha state and is mostly here to enable internal testing\n#       for now. Use with caution.\ninternal-encryption: \"false\"\n\n# internal-trust indicates whether internal traffic is encrypted and mutual trust is achieved.\n# Using internal-trust protect against MITM attacks and ensures that all requests received\n# will go via the ingress and will come from an approved set of sources \n# (e.g. KServ from recognized namespaces)\n# If this is \"true\", the following traffic is encrypted and client authentication is verified with mutual TLS:\n#  - ingress to activator\n#  - ingress to queue-proxy\n#  - activator to queue-proxy\n#\n# NOTE: This flag is in an alpha state and is mostly here to enable internal testing\n#       for now. Use with caution.\ninternal-trust: \"false\"\n"
```
evankanderson, line 27: This got corrupted to containing internal newlines...

pkg/config/config.go
evankanderson: I'm fine with that, but I might say "InternalTrust enforces InternalEncryption = true".

We might also do that in `NewConfigFromMap`.

kvmware:
the yaml lint failure is due to too many fixer warning which will be fixed once this https://github.com/knative/actions/pull/120 lands. The golang linter, I have no idea why it is checking other files not being changed, my local test https://github.com/kvmware/networking/pull/1 does not have that issue.

davidhadas:
> the yaml lint failure is due to too many fixer warning which will be fixed once this [knative/actions#120](https://github.com/knative/actions/pull/120) lands. The golang linter, I have no idea why it is checking other files not being changed, my local test [kvmware#1](https://github.com/kvmware/networking/pull/1) does not have that issue.

I changed the file that is being reported - but the warnings are not in the lines changed. 
Lets wait for the update and hopefully we will be fine

davidhadas:
/hold
See discussion at https://github.com/knative/serving/issues/13764
The plan is to merge this flag with the flag now discussed there.

config/config-network.yaml
```diff
@@ -173,12 +173,25 @@ data:
     # Knative doesn't know about that otherwise.
     default-external-scheme: "http"
 
-    # internal-encryption indicates whether internal traffic is encrypted or not.
-    # If this is "true", the following traffic are encrypted:
-    #  - ingress to activator
-    #  - ingress to queue-proxy
-    #  - activator to queue-proxy
+    # internal-encryption applies to the knative control-plane only
```
evankanderson, line 176: Wait -- it seems like you're re-purposing this flag to do something different?

It seems like this should continue to control the data plane server-authenticated encryption path it does today, but I'm 100% on a better flag going forward using the below.

I'd even be okay with something like:

> Setting this flag to "true" is equivalent to setting `internal-trust` to `minimal`.

You could also say:

> This flag acts as a global disable for the internal-trust feature below.  If this flag is set to "false", the "internal-trust" feature will be disabled.

config/config-network.yaml
```diff
@@ -173,12 +173,25 @@ data:
...
+    #  - ingress to queue-proxy
+    #  - activator to queue-proxy
+    #
+    # internal-trust = "minimal" ensures requests arriving the Kingress will be forwarded to Kserv with a certificate created by the Knative CA
+    # internal-trust = "namespace" ensures requests arriving the Kingress will be forwarded to Kserv with a namespace specific certificate created by the Knative CA
+    # internal-trust = "mutual" same as "namespace" and also ensures that Kserv will only receive reqeusts via the Kingress
+    # internal-trust = "identity" same as "mutual" and also adds sender identity verified by the Kingress and forwarded to Kserv
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing for now. Use with caution.
+    internal-trust: "none"
```
evankanderson, line 197: I'm not sure that "trust" is the right word here -- maybe you mean "internal-auth" or "internal-verification"?  We're _establishing_ trust, but the variable name could also be seen as _declaring_ trust without an enforcement mechanism.

pkg/config/config.go
```diff
@@ -251,8 +255,12 @@ type Config struct {
 	// not enabled. Defaults to "http".
 	DefaultExternalScheme string
 
-	// DefaultExternal specifies whether internal traffic is encrypted or not.
+	// InternalEncryption specifies whether internal traffic is encrypted or not.
 	InternalEncryption bool
+
+	// InternalTrust specifies what level of trust whether internal traffic is encrypted with mutual trust between peers.
+	// InternalTrust enforces InternalEncryption = true
```
evankanderson, line 262: Do we want still want to force InternalEncryption to "true", or do we want this to define additional settings if InternalEncryption is "true"?

pkg/config/config.go
```diff
@@ -295,6 +304,7 @@ func NewConfigFromMap(data map[string]string) (*Config, error) {
 		cm.AsBool(EnableMeshPodAddressabilityKey, &nc.EnableMeshPodAddressability),
 		cm.AsString(DefaultExternalSchemeKey, &nc.DefaultExternalScheme),
 		cm.AsBool(InternalEncryptionKey, &nc.InternalEncryption),
+		cm.AsString(InternalTrustKey, &nc.InternalTrust),
```
evankanderson, line 307: Not your fault, but I wish we had `AsEnum` here.

(Don't take action on this gripe in this PR)

evankanderson (commented):
I'm being picky about names because these are presumably going to be interfaces that administrators will continue to use for years.

I'm 👍  on the general intent of this change at this point.

pkg/config/config.go
davidhadas: missd this - will fix

config/config-network.yaml
davidhadas: > Wait -- it seems like you're re-purposing this flag to do something different?
Yes, this is what I was proposing. 
Before the change, this flag affected both data-plane and control-plane and I am suggesting it will be dedicated to control-plane only. 

> 
> It seems like this should continue to control the data plane server-authenticated encryption path it does today, but I'm 100% on a better flag going forward using the below.
I am proposing that this flag be changed (we are in alpha, so we can still do that) and for data-plane the functionality will be moved to the other flag we now add

config/config-network.yaml
davidhadas: > I'm not sure that "trust" is the right word here -- maybe you mean "internal-auth" or "internal-verification"? We're _establishing_ trust, but the variable name could also be seen as _declaring_ trust without an enforcement mechanism.

It is ok to change trust, but it seems better than auth even verification. 
since it is suggested to control if we use any description and if we do use encryption, then, verifications do we do.

pkg/config/config.go
davidhadas: Removed.

davidhadas:
/unhold
We need to complete this to start making changes to different packages that will now depend on the flags

config/config-network.yaml
davidhadas: The flag is no longer repurposed - it is deprecated and separated to two flags for DP and CP, each with multiple options - this will allow growing the functionality in a gradual non disruptive way

config/config-network.yaml
```diff
@@ -173,12 +173,34 @@ data:
...
+
+    # internal-controlplane-trust indicates the level of trust established in the knative control-plane.
+    # internal-controlplane-trust = "disabled" (the default) - uses no encryption for internal traffic
+    # Using any other value ensures that control traffic is encrypted using TLS.
+    #
+    # internal-controlplane-trust = "enabled" ensures control messages are encrypted using TLS (client authenticate the server)
+    # internal-controlplane-trust = "mutual" ensures control messages are encrypted using mTLS (client and server authenticate each other)
+    #
+    # NOTE: This flag is in an alpha state and is mostly here to enable internal testing for now. Use with caution.
+    internal-controlplane-trust: "disabled"
```
nak3, line 206: I think I understood about `internal-dataplane-trust` but I am still not sure how `internal-controlplane-trust` will be used. Could you share the link docs or github issue for your design/idea where I can refer to?

config/config-network.yaml
```diff
@@ -173,12 +173,34 @@ data:
...
     internal-encryption: "false"
+
+    # internal-dataplane-trust indicates the level of trust established in the knative data-plane.
+    # internal-dataplane-trust = "disabled" (the default) - uses no encryption for internal traffic
+    # Using any other value ensures that the following traffic is encrypted using TLS:
+    #  - ingress to activator
+    #  - ingress to queue-proxy
+    #  - activator to queue-proxy
+    #
+    # internal-dataplane-trust = "minimal" ensures data messages are encrypted, Kingress authenticate that the receiver is a Kserv
```
nak3, line 190: nit: Please change `Kserv` to `Ksvc` or `Kservice`. IMO, `Kserv` or `Kserve` is not used in general and it might be confused with  https://github.com/kserve/kserve

pkg/config/config.go
```diff
@@ -251,8 +260,15 @@ type Config struct {
 	// not enabled. Defaults to "http".
 	DefaultExternalScheme string
 
-	// DefaultExternal specifies whether internal traffic is encrypted or not.
+	// Deprecated - replaced with InternalDataplaneTrust and InternalControlplaneTrust
+	// InternalEncryption specifies whether internal traffic is encrypted or not.
 	InternalEncryption bool
+
+	// InternalDataplaneTrust specifies the level of trust used for date plane.
+	InternalDataplaneTrust string
```
nak3, line 268: Since the values will be referred by other repos, should the values `disabled`, `minimum`, `enabled`, ...etc be introduced as new type like HTTPProtocol https://github.com/knative/networking/blob/68725bdd1056db9dcf7f35b1ea192198045b4351/pkg/config/config.go#L132 ?

pkg/config/config.go
```diff
@@ -295,6 +313,8 @@ func NewConfigFromMap(data map[string]string) (*Config, error) {
 		cm.AsBool(EnableMeshPodAddressabilityKey, &nc.EnableMeshPodAddressability),
 		cm.AsString(DefaultExternalSchemeKey, &nc.DefaultExternalScheme),
 		cm.AsBool(InternalEncryptionKey, &nc.InternalEncryption),
+		cm.AsString(InternalDataplaneTrustKey, &nc.InternalDataplaneTrust),
```
nak3, line 316: Should we add the validation under L#323?

config/config-network.yaml
```diff
@@ -22,9 +22,9 @@ metadata:
     app.kubernetes.io/component: networking
     app.kubernetes.io/version: devel
   annotations:
-    knative.dev/example-checksum: "73d96d1b"
+    knative.dev/example-checksum: "dd928d80"
 data:
-  _example: |
+  _example: |-
```
nak3, line 27: Did you need change? If not, I would like to keep it as it is.

config/config-network.yaml
```diff
@@ -173,12 +173,34 @@ data:
...
-    # If this is "true", the following traffic are encrypted:
-    #  - ingress to activator
-    #  - ingress to queue-proxy
-    #  - activator to queue-proxy
     #
     # NOTE: This flag is in an alpha state and is mostly here to enable internal testing
     #       for now. Use with caution.
     internal-encryption: "false"
+
+    # internal-dataplane-trust indicates the level of trust established in the knative data-plane.
```
nak3, line 183: @davidhadas @evankanderson 
Do we need `internal-` prefix for `internal-controlplane-trust` and `internal-dataplane-trust` ? 
I feel `controlplane-trust` and `dataplane-trust` are enough and they implicate "internal".

config/config-network.yaml
davidhadas: @nak3 
AFAIK, we have little on this right now. 
We understand that we need to start building control plane trust (after we complete the more critical data plane trust) 
And we define a flag that will allow the introduction of such work moving forward.

The little we do have is now summarized in https://docs.google.com/document/d/19R0YOBAZfed2bl99-fdUnQ89cz3hoPY6KJNxhn6b40A/edit#heading=h.n8a530nnrb  - which is incomplete and need more contributions.

pkg/config/config.go
davidhadas: I agree that we should. We would not want a 'disabled' default for cases where someone makes a type and writes 'enbled'. we also cant have 'enabled' anything with tls as default - so the bottom line, we can't have a default which means we need to verify that one of the legal values is used.

PR #787, opened 2023-04-05
Mark `NewConfigFromConfigMap()` as Deprecated in `deprecated_config.go`

nak3:
As per title, `NewConfigFromConfigMap()` is not deprecated even in `deprecated_config.go`
so there is no warning message.

This patch adds the `Deprecated:` tag.

/cc @dprotaso

### DIFF
--- a/pkg/deprecated_config.go
+++ b/pkg/deprecated_config.go
@@ -203,6 +203,8 @@ const (
 )
 
 // NewConfigFromConfigMap creates a Config from the supplied ConfigMap
+//
+// Deprecated: Use knative.dev/networking/pkg/config/NewConfigFromMap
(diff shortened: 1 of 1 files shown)


nak3:
@dprotaso Do you remember why NewConfigFromConfigMap was not added in `pkg/config/config.go` ?

dprotaso:
There's an open issue about reducing the size of the queue proxy. Pulling any k8s deps increases the size by a few MB (maybe 10). So the idea was to split networking into smaller packages.

So I don't recommend moving the function. It may not make sense to deprecate either as there isn't an alternative.

nak3:
Thank you. I see. The file name is `deprecated_config.go` so I was confused.
I am writing a new PR but will use this function (`NewConfigFromConfigMap`), then.

PR #806, opened 2023-05-10
Backward compatibility for InternalEncryption

davidhadas:
Ensure that if someone uses the deprecated InternalEncryption, we use Dataplane-trust=Minimal as a minimum. 
InternalEncryption is deprecated but will take time to remove from dependencies


nak3:
@davidhadas Could you fix the unit test?

nak3:
And this must be merged after https://github.com/knative/serving/pull/13969 ?

davidhadas:
> Ensure that if someone uses the deprecated InternalEncryption, we use Dataplane-trust=Minimal as a minimum.

The idea was to merge before it actually :)

PR #830, opened 2023-07-31
Add dependabot and auto hack scripts

nak3:
As per title, this patch adds dependabot.

/cc @kvmware @dprotaso

### DIFF
--- a/.github/workflows/knative-dependabot.yaml
+++ b/.github/workflows/knative-dependabot.yaml
@@ -1,0 +1,38 @@
+name: dependabot - hack scripts 
+
+on:
+  pull_request:
+
+permissions:
+  contents: write
+
+jobs:
+  generate_some_code:
+    name: update deps
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/knative-dependabot.yaml
```diff
@@ -0,0 +1,38 @@
+name: dependabot - hack scripts
```
kvmware, line 1: This should be added to either actions or .github repo so that we can just sync the file across repos. Not everyone uses dependabot but there are a few that do and would benefit from it.

kvmware:
We already merged this dependabot workflow and it is my bad for not mentioning this then. However, we should generalize this and move it to `actions` or `.github` so others can take advantage.

.github/workflows/knative-dependabot.yaml
nak3: Okay, I was hesitated to adding the common dependabot because some repo has their own dependabot like https://github.com/knative-extensions/eventing-kafka-broker/blob/main/.github/workflows/build-dependabot-pr.yaml
But I will add it to the common one.

PR #831, opened 2023-07-31
Fix comment for old config

nak3:
This patch makes a tiny change which fix a old config name in the comment line.


ReToCode (approved):
/lgtm
/approve

PR #842, opened 2023-08-14
Network Config Changes

KauzClay:
# Changes

- :gift: Add new features

- Adds a function to quickly tell if InternalTLS is enabled ( a common thing to do now that we've added the dataplane-trust flag)
- Adds a function to create a config from a configmap. I've seen this some places and not on others. I'm not sure if this is a pattern we are trying to get away from?
-
-

/kind enhancement

Fixes #

**Release Note**

```release-note
N/A
```

**Docs**

```docs
N/A
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"text/template"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/lru"
 	cm "knative.dev/pkg/configmap"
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -321,6 +322,14 @@ func defaultConfig() *Config {
 	}
 }
 
+// NewConfigFromConfigMap returns a Config for the given configmap
+func NewConfigFromConfigMap(config *corev1.ConfigMap) (*Config, error) {
+	if config == nil {
+		return NewConfigFromMap(nil)
```
nak3, line 328:
```suggestion
                return defaultConfig(), nil
```

It depends on the person but I think it is better to simply return `defaultConfig()`.

nak3: Another code like `leaderelection` has it https://github.com/knative/pkg/blob/main/leaderelection/config.go#L74-L76

nak3 (commented):
/hold
/approve
/lgtm

Just a nit comment so please feel free to unhold if you will not change it.

pkg/config/config.go
KauzClay: @nak3 

hmm interesting, since the beginning of NewConfigFromMap uses defaultConfig anyway, why not just skip all the extra stuff and return it directly?

I think I see your point. However, for now I will leave it as is, my reason being:

There are quite a few different checks in the NewConfigFromMap, and I feel safer having then run no matter if there is a configmap or not. Then, on the chance we change something about the default config, we have more code checks in place to yell at us if we end up changing a behavior we weren't expecting.

However, I'm happy to hear your response, I can always open a new PR later if you don't think my concern would actually be a problem 👍

PR #846, opened 2023-08-28
Fix inaccurate instruction for internal-encryption

nak3:
Fix the comment in the config map as neither `internal-dataplane-trust` nor `internal-controlplane-trust` exists.

**Release Note**

```release-note
NONE
```


ReToCode (approved):
/lgtm

PR #847, opened 2023-08-28
Improve InternalTLSEnabled util function

nak3:
Currently `InternalTLSEnabled()` has two issues such as:

* `true` is returned when `dataplane-trust` is not configured. (The current default is `disabled`.)
* ~`controlplane-trust` is not considered.~

This patch improves these issues.

/kind enhancement

**Release Note**

```release-note
NONE
```


ReToCode (approved):
/lgtm

PR #861, opened 2023-09-18
Cleanup SAN constants and Secrets Keys for system-internal-tls certificates

ReToCode:
# Changes
See https://github.com/knative/serving/issues/14392#issue-1900760814

/kind cleanup

Partially https://github.com/knative/serving/issues/14392


ReToCode:
Basically the same applies for CI as in https://github.com/knative/networking/pull/858#issuecomment-1719385001.
We are half-using this this stuff and half-not using this stuff, so it's a bit hard to clean it up without breaking.

/assign @nak3 
/assign @KauzClay 
/assign @dprotaso

ReToCode:
Ah, forget the comment about CI. Seems we did not even us any of those flags just yet.

pkg/certificates/constants.go
```diff
@@ -20,41 +20,39 @@ import "strings"
...
-	ControlPlaneName       = "kn-control"
-
-	//These keys are meant to line up with cert-manager, see
-	//https://cert-manager.io/docs/usage/certificate/#additional-certificate-output-formats
+
+	// nolint:all
+	// Deprecated: FakeDnsName is deprecated.
+	// Please use the DataPlaneRoutingSAN for calls to the Activator
+	// and the DataPlaneUserSAN function for calls to a Knative-Service via Queue-Proxy.
+	FakeDnsName = LegacyFakeDnsName
```
dprotaso, line 31:
```suggestion
	FakeDNSName = LegacyFakeDnsName
```

Technically we should be casing this properly 

and a bunch of other properties below

ReToCode: Agreed, but these are already used in this form. 
If we want to fix that, we probably need to create more reverse compatible deprecated fields.

dprotaso: I guess we're going to drop this field soon

PR #863, opened 2023-09-21
Verify and reconcile new keys' value instead of obsleted keys' in secret

nak3:
### Changes

https://github.com/knative-extensions/control-protocol/pull/220 introduced new keys and https://github.com/knative/serving/pull/13388 adopted them but the new keys are not verified or reconciled via reconciler.

This patch changes the new keys' values to be reconciled.

### Alternative

Alternatively, we can reconcile both new keys' values and old keys' values but I think we can drop the old keys now in the follow up PR.

### Additional info

Please refer to https://github.com/knative/serving/issues/14407 for the reproducer.

Fix https://github.com/knative/serving/issues/14407

### DIFF
--- a/pkg/certificates/reconciler/certificates.go
+++ b/pkg/certificates/reconciler/certificates.go
@@ -119,7 +119,7 @@ func (r *reconciler) ReconcileKind(ctx context.Context, secret *corev1.Secret) p
 		return fmt.Errorf("unknown cert type: %v", r.secretTypeLabelName)
 	}
 
-	cert, _, err := parseAndValidateSecret(secret, caSecret.Data[certificates.SecretCertKey], sans...)
+	cert, _, err := parseAndValidateSecret(secret, caSecret.Data[certificates.CertName], sans...)
 	if err != nil {
 		r.logger.Infof("Secret invalid: %v", err)
 		// Check the secret to reconcile type
@@ -129,7 +129,7 @@ func (r *reconciler) ReconcileKind(ctx context.Context, secret *corev1.Secret) p
 		if err != nil {
 			return fmt.Errorf("cannot generate the cert: %w", err)
 		}
-		err = r.commitUpdatedSecret(ctx, secret, keyPair, caSecret.Data[certificates.SecretCertKey])
+		err = r.commitUpdatedSecret(ctx, secret, keyPair, caSecret.Data[certificates.CertName])
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/certificates/reconciler/certificates.go
```diff
@@ -146,21 +146,21 @@ func (r *reconciler) ReconcileKind(ctx context.Context, secret *corev1.Secret) p
...
 	if caCert != nil {
-		ca, ok := secret.Data[certificates.SecretCaCertKey]
+		ca, ok := secret.Data[certificates.CaCertName]
 		if !ok {
-			return nil, nil, fmt.Errorf("missing ca cert bytes")
+			return nil, nil, fmt.Errorf("missing ca cert bytes in %q", certificates.CaCertName)
 		}
 		if !bytes.Equal(ca, caCert) {
-			return nil, nil, fmt.Errorf("ca cert bytes changed")
+			return nil, nil, fmt.Errorf("ca cert bytes changed in %q", certificates.CaCertName)
```
nak3, line 163: Note, updating these log because the secret may contain both `ca-cert.pem` and `ca.crt` so it is difficult for users to understand whether controller complains old key's value or new keys' value.

pkg/certificates/reconciler/certificates.go
```diff
@@ -146,21 +146,21 @@ func (r *reconciler) ReconcileKind(ctx context.Context, secret *corev1.Secret) p
 
 // All sans provided are required to be lower case
 func parseAndValidateSecret(secret *corev1.Secret, caCert []byte, sans ...string) (*x509.Certificate, *rsa.PrivateKey, error) {
-	certBytes, ok := secret.Data[certificates.SecretCertKey]
+	certBytes, ok := secret.Data[certificates.CertName]
 	if !ok {
-		return nil, nil, fmt.Errorf("missing cert bytes")
+		return nil, nil, fmt.Errorf("missing cert bytes in %q", certificates.CertName)
```
ReToCode, line 151: can we also add the secret name + namespace?

ReToCode (commented):
Other than that one log, LGTM.

pkg/certificates/reconciler/certificates.go
nak3: Yes, current prints the logs like:

```
{"severity":"INFO","timestamp":"2023-09-21T06:09:48.57517051Z","logger":"controller","caller":"reconciler/certificates.go:80","message":"Updating secret \"knative-serving\":\"knative-serving-certs\"","commit":"28effd9","knative.dev/pod":"controller-b98b48949-88w7f"}
{"severity":"INFO","timestamp":"2023-09-21T06:09:48.575482762Z","logger":"controller","caller":"reconciler/certificates.go:124","message":"Secret invalid: missing ca cert bytes","commit":"28effd9","knative.dev/pod":"controller-b98b48949-88w7f"}
```

So we should add it. I added the log in L124 above with the common `namespace/name` pattern.

ReToCode (approved):
/lgtm
/approve

PR #904, opened 2024-01-03
add helper for determining app protocol

KauzClay:
# Changes

- :gift: new helper function for determining AppProtocol value

/kind enhancement

In support of https://github.com/knative/serving/issues/14569

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/apis/networking/ports.go
+++ b/pkg/apis/networking/ports.go
@@ -38,6 +38,9 @@ const (
 
 	// ServicePortNameHTTPS is the name of the external port of the service for HTTPS
 	ServicePortNameHTTPS = "https"
+
+	// AppProtocolH2C is the name of the external port of the service for HTTP/2
+	AppProtocolH2C = "kubernetes.io/h2c"
 )
 
 // ServicePortName returns the port for the app level protocol.
@@ -54,4 +57,12 @@ func ServicePort(proto ProtocolType) int {
 		return ServiceHTTP2Port
 	}
 	return ServiceHTTPPort
+}
+
+// AppProtocol returns the value for app level protocol based on the ProtocolType
+func AppProtocol(proto ProtocolType) string {
+	if proto == ProtocolH2C {
+		return AppProtocolH2C
+	}
+	return ""
 }


KauzClay:
Question for reviewers: I ended up creating the variable `AppProtocolH2C` as a var instead of a const, because I needed to use it in a string pointer. My understanding in go is that you can't take a pointer to a const.

But, given k8s uses string pointers for other stuff, is there maybe a nicer way to do this that I'm not aware of?

pkg/apis/networking/ports.go
```diff
@@ -40,6 +40,11 @@ const (
 	ServicePortNameHTTPS = "https"
 )
 
+var (
+	// AppProtocolH2C is the name of the external port of the service for HTTP/2
+	AppProtocolH2C = "kubernetes.io/h2c"
```
skonto, line 45: How about `kubernetes.io/ws` ?

skonto: Could we somehow make clear that this is gateway api related?

KauzClay: > How about kubernetes.io/ws? Is there a plan for this?

Ah, good point. I think I may have misinterpretted Dave's sentence on WS here: https://github.com/knative/serving/issues/14569


He says:
> For websocket - this in theory should just be on by default (contour toggled it on for Gateway API). There might be other implementations that have this off but let's revisit that when we find one that doesn't

Which when I first read it, I thought, leave it and come back to it, but perhaps what he's really saying is put it on always?

KauzClay: > Could we somehow make clear that this is gateway api related?

Sure, like perhaps calling it `GatewayAPIAppProtocolH2C` or something?

KauzClay: ah, you know, reading [Dave's GEP](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1911/index.md#default-protocols), maybe we don't need to do anything for WS, since upgrades happen by default?

@dprotaso what were you imagining here?

also, how does one indicate that a Knative Service should be a websocket?

dprotaso: > Sure, like perhaps calling it GatewayAPIAppProtocolH2C or something?

These constants are technically Kubernetes constants - I don't think we should prefix them with the Gateway API

We can link the KEP so it's clear

https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols#new-standard-protocols

Honestly, websocket is a weird one cause Contour should have probably not disabled it. For now we can skip it since I'd imagine most implementations support the upgrade flow via HTTP

So we really need to set the h2c protocol when the user specifies it in their port name (the way it's used for grpc services)

eg.
https://github.com/knative/docs/blob/dab74765e0f0c60789fe97208aa0b0a1e23eb9cd/code-samples/serving/grpc-ping-go/service.yaml#L12C13-L12C22

dprotaso: If we need websocket support we would have to upgrade the runtime contract to define a constants users would use

https://github.com/knative/specs/blob/main/specs/serving/runtime-contract.md#protocols-and-ports

pkg/apis/networking/ports_test.go
```diff
@@ -65,3 +65,26 @@ func TestServicePort(t *testing.T) {
...
+		proto:  ProtocolH2C,
+		expect: &AppProtocolH2C,
+	}, {
+		name:   "other protocols result in nil",
+		proto:  ProtocolHTTP1,
+		expect: nil,
+	}}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if got, want := AppProtocol(c.proto), c.expect; !(got == want) {
```
houshengbo, line 85: This is something trivial, but you can use `got != want`, instead of `!(got == want)`.

pkg/apis/networking/ports.go
KauzClay: so, to summarize:
* keep the variable name, but link the KEP in the comment
* don't move forward with WS right now

is that correct?

As far as this comment:
> So we really need to set the h2c protocol when the user specifies it in their port name (the way it's used for grpc services)

My other PR (https://github.com/knative/serving/pull/14757) is where that would happen, right?

dprotaso: > My other PR (knative/serving#14757) is where that would happen, right?

Yeah

PR #908, opened 2024-01-09
Replace deprecated `sets` usages with generic version

ReToCode:
# Changes
- Follow-up for https://github.com/knative/pkg/pull/2915
- Replaces all deprecated usages of `sets` with new generic versions

/kind cleanup

Partially https://github.com/knative/serving/issues/14692

### DIFF
--- a/pkg/apis/networking/metadata_validation.go
+++ b/pkg/apis/networking/metadata_validation.go
@@ -25,7 +25,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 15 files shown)


ReToCode:
@dprotaso @skonto as we now also have https://github.com/knative/pkg/pull/2915 merged, we need to update networking, Serving and some net-* repos to make deps bumps work again.

PR #926, opened 2024-02-18
feat: add isFailed for certificate

ckcd:
# Changes

- :gift: Add `isFailed` lifecycle method for `Certificate`.

/kind enhancement

In support of #https://github.com/knative/serving/issues/14530
And after this we need to make an adaptation in `Serving`.

**Release Note**

```release-note
N/A
```

**Docs**

```docs
N/A
```

### DIFF
--- a/pkg/apis/networking/v1alpha1/certificate_lifecycle.go
+++ b/pkg/apis/networking/v1alpha1/certificate_lifecycle.go
@@ -58,6 +58,14 @@ func (c *Certificate) IsReady() bool {
 		cs.GetCondition(CertificateConditionReady).IsTrue()
 }
 
+// IsFailed returns true is the Certificate ready is false
+// and the Certificate resource has been observed.
+func (c *Certificate) IsFailed() bool {
+	cs := c.Status
+	return cs.ObservedGeneration == c.Generation &&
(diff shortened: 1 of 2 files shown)


ckcd:
@gabo1208 @ReToCode PTAL, thanks.

pkg/apis/networking/v1alpha1/certificate_lifecycle.go
```diff
@@ -58,6 +58,14 @@ func (c *Certificate) IsReady() bool {
 		cs.GetCondition(CertificateConditionReady).IsTrue()
 }
 
+// IsFailed returns true is the Certificate ready is false
+// and the Certificate resource has been observed.
+func (c *Certificate) IsFailed() bool {
```
ReToCode, line 63: Do we need that extra method? Could the user not just use `!IsReady`?

ckcd: @ReToCode Thanks for your reply. 
I think `IdReady` is not enough because the `ConditionStatus` has three possible value: `true`, `false` or `unknown`. `IsReady` only checks whether it's `true`. And the new `IsFailed` check whether it's `false`.

ReToCode (approved):
/lgtm
/approve

PR #932, opened 2024-02-22
Update grpc/h2c conformance test 

dprotaso:
Part of https://github.com/knative/serving/issues/14569

As part of the Gateway API work backend protocols can be
hinted using a Service's `appProtocol` attribute.

This sets the attribute on the h2c/grpc test

See the GEP: https://gateway-api.sigs.k8s.io/geps/gep-1911/


dprotaso:
The failure is fixed in https://github.com/knative/networking/pull/931

dprotaso:
Re-opening the PR supposedly updates the target branches's SHA - so the re-running of the actions should pick up latest changess in `main`

PR #933, opened 2024-02-22
drop unused types

dprotaso:
These types were added to support a feature track defined here - https://docs.google.com/document/d/1kefwbDQ6qCsw-DbqmfpLlQfEvLtO85RgUlQMCqJnTLg/edit?resourcekey=0-lzesL79SETekPVyzEfc4Ww#heading=h.n8a530nnrb

This seems less relevant with Gateway API and the feature track has been abandoned for a few years


dprotaso:
codecov is broken in this repo

ReToCode (approved):
/lgtm
/approve

skonto:
>This seems less relevant with Gateway API 

Is there are a way to achieve the same with the old design doc?

dprotaso:
/hold cancel

> Is there are a way to achieve the same with the old design doc?

I think it would require some work on us to coordinate domain selection and gateway selection. This might be doable without CRDs and just operator knobs in our config maps - eg. see the istio gateway selection feature track

PR #950, opened 2024-03-25
Drop special logic for kubelet probe header for istio

izabelacg:
# Changes

- Deprecates k-kubelet-probe header

Part of https://github.com/knative/serving/issues/14981

### DIFF
--- a/pkg/deprecated_header.go
+++ b/pkg/deprecated_header.go
@@ -66,15 +66,8 @@ const (
 	// probes.  Istio with mTLS rewrites probes, but their probes pass a
 	// different user-agent.  So we augment the probes with this header.
 	//
-	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
-	KubeletProbeHeaderName = header.UserAgentKey
-
-	// KubeletProbeKey is the name of the header supplied by kubelet
-	// probes.  Istio with mTLS rewrites probes, but their probes pass a
-	// different user-agent.  So we augment the probes with this header.
-	//
-	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
-	KubeletProbeKey = header.UserAgentKey
+	// Deprecated: use knative.dev/networking/pkg/http/header.KubeletProbeKey
+	KubeletProbeHeaderName = header.KubeletProbeKey
 
 	// UserAgentKey is the constant for header "User-Agent".
 	//
--- a/pkg/deprecated_header.go
+++ b/pkg/deprecated_header.go
@@ -66,8 +66,15 @@ const (
 	// probes.  Istio with mTLS rewrites probes, but their probes pass a
 	// different user-agent.  So we augment the probes with this header.
 	//
-	// Deprecated: use knative.dev/networking/pkg/http/header.KubeletProbeKey
-	KubeletProbeHeaderName = header.KubeletProbeKey
+	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
+	KubeletProbeHeaderName = header.UserAgentKey
+
+	// KubeletProbeKey is the name of the header supplied by kubelet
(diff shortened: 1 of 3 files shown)


izabelacg:
The downstream test TestIgnoreProbe fails since it is setting k-kubelet-probe to signal a kubelet probe, however the logic to determine a kubelet probe is now different and no longer checks for the custom header.

dprotaso:
These constants are used in knative/pkg - we do not want to deprecate them

see: https://github.com/knative/pkg/pull/2997#issuecomment-2020368239

dprotaso:
Whoops my bad - for some reason I thought `K-Kubelet-Probe` was a K8s thing - but they just use `User-Agent` with the `kube-probe/` prefix

pkg/deprecated_header.go
```diff
@@ -66,8 +66,15 @@ const (
...
-	KubeletProbeHeaderName = header.KubeletProbeKey
+	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
+	KubeletProbeHeaderName = header.UserAgentKey
+
+	// KubeletProbeKey is the name of the header supplied by kubelet
+	// probes.  Istio with mTLS rewrites probes, but their probes pass a
+	// different user-agent.  So we augment the probes with this header.
+	//
+	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
+	KubeletProbeKey = header.UserAgentKey
```
dprotaso, line 77: Drop this addtion - we don't need to introduce a new field here

pkg/deprecated_header.go
```diff
@@ -66,8 +66,15 @@ const (
 	// probes.  Istio with mTLS rewrites probes, but their probes pass a
 	// different user-agent.  So we augment the probes with this header.
 	//
-	// Deprecated: use knative.dev/networking/pkg/http/header.KubeletProbeKey
-	KubeletProbeHeaderName = header.KubeletProbeKey
+	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
+	KubeletProbeHeaderName = header.UserAgentKey
```
dprotaso, line 70: I'd leave this referencing `header.KubeletProbeKey` as before

pkg/http/header/header.go
```diff
@@ -58,6 +58,8 @@ const (
 	// KubeletProbeKey is the name of the header supplied by kubelet
 	// probes.  Istio with mTLS rewrites probes, but their probes pass a
 	// different user-agent.  So we augment the probes with this header.
+	//
+	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
```
dprotaso, line 62: I'd be more specific in this deprecation notice and say we want to detect probes with `header.UserAgentKey` and `KubeProbeUAPrefix`

pkg/http/header/header.go
```diff
@@ -126,8 +128,7 @@ func IsProbe(r *http.Request) bool {
 
 // IsKubeletProbe returns true if the request is a Kubernetes probe.
 func IsKubeletProbe(r *http.Request) bool {
-	return strings.HasPrefix(r.Header.Get("User-Agent"), KubeProbeUAPrefix) ||
-		r.Header.Get(KubeletProbeKey) != ""
+	return strings.HasPrefix(r.Header.Get("User-Agent"), KubeProbeUAPrefix)
```
dprotaso, line 131: I think we should leave this as is for now and then remove the use of `K-Kubelet-Probe` after a release or two

dprotaso: if there's a minor drift between the queue proxy and the deployment then we want to prevent things from blowing up.

pkg/http/header/header.go
izabelacg: Made changes. Let me know if the text makes sense

PR #957, opened 2024-04-02
Replace all usages of deprecated wait.PollImmediate with wait.PollUntilContextTimeout

prashantrewar:
# Changes

- Replace usages of deprecated wait.PollImmediate with wait.PollUntilContextTimeout

/kind <kind>

Fixes #

**Release Note**

```release-note
NONE
```

**Docs**

```docs

```

### DIFF
--- a/pkg/prober/prober.go
+++ b/pkg/prober/prober.go
@@ -186,7 +186,7 @@ func (m *Manager) doAsync(ctx context.Context, target string, arg interface{}, p
 			result bool
 			inErr  error
 		)
-		err := wait.PollImmediate(period, timeout, func() (bool, error) {
+		err := wait.PollUntilContextTimeout(ctx, period, timeout, true, func(ctx context.Context) (bool, error) {
 			result, inErr = Do(ctx, m.transport, target, ops...)
 			// Do not return error, which is from verifierError, as retry is expected until timeout.
 			return result, nil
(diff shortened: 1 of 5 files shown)


dprotaso:
I believe the test is failing because `wait.ErrWaitTimeout` is not longer returned by these new functions

https://github.com/knative/networking/blob/13c6ea7211e1702121c36cd88c3c049e4eeb6515/pkg/prober/prober_test.go#L227-L229


According to the docs https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Poll
> Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.

dprotaso:
You'll also need to verify if `wait.ErrWaitTimeout` is used anywhere in the serving/net- repo and update those accordingly.

dprotaso:
I don't think it's used anywhere in code that isn't tests so I think we should be ok

ReToCode:
@prashantrewar mind taking a look at the test failures?

dprotaso:
@prashantrewar just following up - are you able to address the unit test failures?

dprotaso:
> Should I replace this also?
Yeah update the test

prashantrewar:
> > Should I replace this also?
> 
> Yeah update the test

@dprotaso Done!

dprotaso:
/lgtm
/approve

thanks @prashantrewar

prashantrewar:
Thank you @dprotaso for guidance and support.

PR #963, opened 2024-04-10
this fixes the probe handler to reflect that it doesn't care about the path

dprotaso:
While working on https://github.com/knative-extensions/net-gateway-api/issues/18

I discovered the runtime image in this repository doesn't behave like the queue proxy. The probe handler should work with _any_ URL path but this image it only works when the path is `/healthz`

/assign @izabelacg

### DIFF
--- a/test/test_images/runtime/handlers/handler.go
+++ b/test/test_images/runtime/handlers/handler.go
@@ -24,16 +24,13 @@ import (
 	"strings"
 
 	nethttp "knative.dev/networking/pkg/http"
-	"knative.dev/networking/pkg/http/probe"
 	"knative.dev/pkg/network"
 )
 
 // InitHandlers initializes all handlers.
 func InitHandlers(mux *http.ServeMux) {
 	mux.HandleFunc("/", withHeaders(withRequestLog(runtimeHandler)))
-
-	h := probe.NewHandler(withRequestLog(withKubeletProbeHeaderCheck))
-	mux.HandleFunc(nethttp.HealthCheckPath, h.ServeHTTP)
+	mux.HandleFunc(nethttp.HealthCheckPath, withRequestLog(withKubeletProbeHeaderCheck))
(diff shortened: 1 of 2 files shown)


dprotaso:
cc @ReToCode @skonto

izabelacg:
can you share how to reproduce the issue? I curled the service (original) in different paths with the user-agent header and seems to work as expected. I get the following:
```
$ curl http://runtime-test-image.default.svc.cluster.local -H "User-Agent: kube-probe/1.25" -v
* Host runtime-test-image.default.svc.cluster.local:80 was resolved.r-Agent: kube-probe/1.25" -v
...
* Connected to runtime-test-image.default.svc.cluster.local (10.96.102.30) port 80
> GET / HTTP/1.1
> Host: runtime-test-image.default.svc.cluster.local
> Accept: */*
> User-Agent: kube-probe/1.25
> 
* Request completely sent off
< HTTP/1.1 200 OK
< date: Wed, 10 Apr 2024 18:08:21 GMT
< content-length: 0
< x-envoy-upstream-service-time: 0
< server: envoy
< 
* Connection #0 to host runtime-test-image.default.svc.cluster.local left intact
```

dprotaso:
When you curl with the `probe` header it should always return the `hash`

eg.

```
curl -vv -H"K-Network-Hash: some-hash" -H"K-Network-Probe: probe" localhost:8080/healthz
```

Should also work when you are not using the `/healthz` path 

```
curl -vv -H"K-Network-Hash: some-hash" -H"K-Network-Probe: probe" localhost:8080
```

dprotaso:
If you're running this image with Knative Service then you won't reproduce original issue

These pods don't run with the queue proxy sidecar in the conformance tests - so it's replicating that.

izabelacg:
Thanks for the explanation. I got to see the issue and solution working.

PR #964, opened 2024-04-10
some conformance tests assume we're getting back a runtime request

dprotaso:
The changes in https://github.com/knative/networking/pull/963 broke some conformance tests.

Prior TestProbeHeaders would 'perform' a runtime request. This would have the image return the headers of the request etc.

Because the handlers were changes in the other PR this runtime info was not being returned causing the test to fail.

We can actually have the test inspect the response headers instead of the json payload.

### DIFF
--- a/test/conformance/ingress/headers.go
+++ b/test/conformance/ingress/headers.go
@@ -83,21 +83,25 @@ func TestProbeHeaders(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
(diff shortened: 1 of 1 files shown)


dprotaso:
/hold cancel 

downstream test passed
> PASS test/conformance.TestIngressConformance/headers/probe/kingress_generates_hash (0.01s)
> PASS test/conformance.TestIngressConformance/headers/probe/request_overrides_hash (0.01s)

PR #970, opened 2024-04-18
Update test dialer to use ip when available

andrew-su:
With the changes to https://github.com/knative-extensions/net-gateway-api/pull/685 where it is possible to have the ingress's public loadbalancer point to an IP instead of domain, we should use the value provided.

# Changes
- :bug: Don't expect there to only be domains, sometimes it might be an IP

/kind bug
/king cleanup

### DIFF
--- a/test/conformance/ingress/util.go
+++ b/test/conformance/ingress/util.go
@@ -1096,10 +1096,8 @@ func CreateDialContext(ctx context.Context, t *testing.T, ing *v1alpha1.Ingress,
 				}
 				return nil, errors.New("service ingress does not contain dialing information")
 			}
-		} else {
-			t.Fatal("Service does not have a supported shape (not type LoadBalancer? missing --ingressendpoint?).")
-			return nil // Unreachable
 		}
+		t.Fatal("Service does not have a supported shape (not type LoadBalancer? missing --ingressendpoint?).")
 	} else if ing.Status.PublicLoadBalancer.Ingress[0].IP != "" {
 		dial := network.NewBackoffDialer(dialBackoff)
 		ingressIP := ing.Status.PublicLoadBalancer.Ingress[0].IP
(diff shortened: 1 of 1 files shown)


test/conformance/ingress/util.go
```diff
@@ -1043,6 +1043,20 @@ func CreateDialContext(ctx context.Context, t *testing.T, ing *v1alpha1.Ingress,
...
+		port := 80
+		if ing.Spec.Rules[0].Visibility == v1alpha1.IngressVisibilityExternalIP && ing.Spec.HTTPOption == v1alpha1.HTTPOptionRedirected {
+			port = 443
+		}
+
+		return func(ctx context.Context, _ string, _ string) (net.Conn, error) {
+			return dial(ctx, "tcp", fmt.Sprintf("%s:%d", ingressIP, port))
+		}
+	}
+
```
dprotaso, line 1059: Can we do this as a fallback when DomainInternal isn't present instead of this being the norm

dprotaso: eg. with net-contour we always return the IP

izabelacg:
@dprotaso is it ok to include this PR (once is merged) in the already created release-1.14 branch?

dprotaso:
Yeah I think it's fine to cherry pick once this merges since it's just some networking conformance test code

test/conformance/ingress/util.go
andrew-su: Updated

dprotaso:
codegen timedout hitting go module cache 

/retest

dprotaso:
@andrew-su see the linter error

dprotaso:
the linter is still going off

dprotaso:
/lgtm
/approve

@izabelacg feel free to cherry-pick

PR #972, opened 2024-04-19
Remove requestDelay flag for tests

andrew-su:
# Changes

- :broom: Remove test flag `requestDelay` as this can cause dispatch test to run for an extremely long time.

/kind cleanup


izabelacg:
@dprotaso is it ok to cherrypick this one to release-1.14 as well?

dprotaso:
No - that'll create churn and there's no utility in merging it back since the behaviour is opt-in

PR #1001, opened 2024-08-20
Make maximum delay of prober in its backoff configurable

SaschaSchwarze0:
# Changes

We are running a larger installation with Knative Serving + Istio + NetIstio. What we observe is that the more services there are, the longer it takes to provision a new KService. This is fine, but at some point, we observe that times increase a lot and containing jumps. E. g. we measure many values around 21 seconds and around 35 seconds but rarely in between. The image is always available so the time it takes to get the Configuration ready is quite stable. The differences we see in the Route.

We think that this has to do with the prober which is containing an exponential back-off with a maximum delay of 30 seconds.

The following table shows how long the prober takes depending on how many tries it needs ignoring the time it actually needs to perform the probe itself = just summing up the delays:

 Try | BackOff | Sum
---- | ------- | ------
   0 |       0 | 0
   1 |    0.05 | 0.05
   2 |     0.1 | 0.15
   3 |     0.2 | 0.35
   4 |     0.4 | 0.75
   5 |     0.8 | 1.55
   6 |     1.6 | 3.15
   7 |     3.2 | 6.35
   8 |     6.4 | 12.75
   9 |    12.8 | 25.55
  10 |    25.6 | 51.15
  11 |      30 | 81.15
  12 |      30 | 111.15

Those numbers very well match what we observe.  For example for those provisions that took roughly 21 seconds, it needed eight probes (12.75 s) and when it needed nine probes (25.55 s) the overall provisioning took around 35 seconds. And yes, we also see provisions that need a little over a minute where it probably needed ten tries to probe successfully.

-- 

While we obviously spend time in improving this with Istio tuning, we would also like to play with lower values for the maximum delay as we think that higher delays (5 seconds and more) are reached too early = already after 6.35 seconds, and compared to the overall load in our system, the number of probes is so minimal that we think that more failed probes would not cause any harm.

This PR therefore introduces a `PROBE_MAX_RETRY_DELAY_SECONDS` environment variable that - if set - will be used as maximum delay.

If you think that makes no sense as a general change, then just close it. If you think it should be exposed in another way, let me know.

/kind enhancement

**Release Note**

```release-note
You can now set a PROBE_MAX_RETRY_DELAY_SECONDS environment variable on your networking component to define a custom maximum retry delay of the prober that will be used instead of the default of 30 seconds.
```

### DIFF
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -23,8 +23,10 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"path"
 	"reflect"
+	"strconv"
 	"sync"
 	"time"
 
@@ -55,6 +57,19 @@ const (
 	// It gives times for the change to propagate and prevents unnecessary retries.
 	initialDelay = 200 * time.Millisecond
 )
+
+var (
+	// probeMaxRetryDelay defines the maximum delay between retries in the backoff of probing
+	probeMaxRetryDelay = 30 * time.Second
+)
+
+func init() {
+	if val, ok := os.LookupEnv("PROBE_MAX_RETRY_DELAY_SECONDS"); ok {
+		if durationSeconds, err := strconv.Atoi(val); err == nil && durationSeconds > 0 {
+			probeMaxRetryDelay = time.Duration(durationSeconds) * time.Second
+		}
+	}
+}
 
 var dialContext = (&net.Dialer{Timeout: probeTimeout}).DialContext
 
@@ -144,7 +159,7 @@ func NewProber(
 		workQueue: workqueue.NewNamedRateLimitingQueue(
 			workqueue.NewMaxOfRateLimiter(
 				// Per item exponential backoff
-				workqueue.NewItemExponentialFailureRateLimiter(50*time.Millisecond, 30*time.Second),
+				workqueue.NewItemExponentialFailureRateLimiter(50*time.Millisecond, probeMaxRetryDelay),
 				// Global rate limiter
 				&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(50), 100)},
 			),


pkg/status/status.go
```diff
@@ -56,6 +58,19 @@ const (
 	initialDelay = 200 * time.Millisecond
 )
 
+var (
+	// probeMaxRetryDelay defines the maximum delay between retries in the backoff of probing
+	probeMaxRetryDelay = 30 * time.Second
+)
+
+func init() {
+	if val, ok := os.LookupEnv("PROBE_MAX_RETRY_DELAY_SECONDS"); ok {
```
skonto, line 67: Should we document this? We could add it at the controller deployment config as well after: 
```
        - name: CONFIG_LOGGING_NAME
          value: config-logging
        - name: CONFIG_OBSERVABILITY_NAME
          value: config-observability
        - name: ENABLE_SECRET_INFORMER_FILTERING_BY_CERT_UID
          value: "false"
```

skonto: Do you have any numbers about the correct value per number of services? That would be also interesting to document as a recommendation.

skonto:
@SaschaSchwarze0  hi, thank you for the PR.

> If you think that makes no sense as a general change, then just close it. If you think it should be exposed in another way, let me know.

@dprotaso any objections on this one? From a tech pov if it solves a problem I am fine, at the end of the day the default remains the same.

> We think that this has to do with the prober which is containing an exponential back-off with a maximum delay of 30 seconds.

 It would be nice to know how different values affect the large deployment to have at least some evidence.

pkg/status/status.go
SaschaSchwarze0: I am not sure if this number should be related to the number of services. The number of services imo has an impact on how long - in our case - Istio will need to get the config synced to all Ingress gateways - and also depends on things like how Istio is tuned (like CPU for istiod), whether delta pushes are enabled for some parts or not, how large the mesh is over for example by using ambient or not etc.

The number we look at here is the maximum duration of the back-off in net-istio. As with exponential back-offs in the context of retries, I assume this was done having in mind to not overload things.

And here is where we at least do not see an overload for us. The two involved components are the net-istio-controller itself that originates the requests. Would for whetever reason be there many KIngresses that it reconciles and then must probe, then yes, reducing the probing max delay increases the load. But in our experience, that is rarely the case and has never been something near to a bottleneck. On the receiving side, the probe requests to the Istio ingress gateway. And there the percentage of requests that are actually probing requests are lik 0.0...1 % ( I have no exact number that I now looked at, but it is so low that it has no relevance at all).

SaschaSchwarze0: > What about other ingresses are they equally affected? 🤔

I have no experience on them.

And yes, adding it to the deployment yaml can be done as a follow-on pr.

SaschaSchwarze0:
@dprotaso @skonto just noticed this did not move forward. Anything else you need from me?

skonto:
@SaschaSchwarze0 hi could you do the follow up PR at the net-istio side?

PR #1029, opened 2025-01-09
drop unnecessary external name svc

dprotaso:
When debugging the visibility split issues in https://github.com/knative-extensions/net-contour/pull/1149 I realize we don't need the extra ExternalName K8s Service.

It was being used to supply a DNS entry that would map a host name to the contour internal envoy service. 

As an alternative we can add a `GATEWAY_HOST` env var that will configure the conformance httpproxy app to just dial the envoy service directly.

### DIFF
--- a/test/conformance/ingress/util.go
+++ b/test/conformance/ingress/util.go
@@ -256,6 +256,9 @@ func CreateProxyService(ctx context.Context, t *testing.T, clients *test.Clients
 					Name:  "TARGET_HOST",
 					Value: target,
 				}, {
+					Name:  "GATEWAY_HOST",
+					Value: gatewayDomain,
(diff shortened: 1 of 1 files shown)


dprotaso:
note: I tested the change in a wip commit here: https://github.com/knative-extensions/net-contour/commit/c0ebfc890ba9bdf5b94ffcf46e6c7f57e7dcbd5b

it passed conformance tests here: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-extensions_net-contour/1149/integration-tests_net-contour_main/1877181980768997376

dprotaso:
For kourier/istio should be fine. I think it should be fine for mesh as well.

PR #1034, opened 2025-01-10
linter fixes

dprotaso:
Linter fixes on https://github.com/knative/networking/pull/1033


dprotaso:
contour failure is because the deps are pinned - we can unpin them after this PR merges - https://github.com/knative-extensions/net-contour/pull/1149

skonto:
>  contour failure is because the deps are pinned - we can unpin them after this PR merges 

Which deps?

dprotaso:
net-contour downstream test this will fail until it has newer k8s libs

dprotaso:
this merged due to this bug - https://github.com/kubernetes-sigs/prow/issues/356

dprotaso:
or wait it passed

PR #1042, opened 2025-01-27
fix codegen by using newer scripts

dprotaso:


dprotaso:
/assign @skonto 

Note: this is targeting main branch

